Issue #81, opened 2020-07-31
Exec format error with bombardier job


jmezach:
I've been trying out crank on my machine to do some benchmarks on a project. I used the sample from the docs and then tweaked it slightly to match my scenario, but I'm only seeing a single request going to my app. Looking at the crank-agent logs I noticed this:

```
[09:02:18.802] > bombardier  -d 15s -c 256 --insecure -l --fasthttp http://localhost:5000/realization --print r --format json
[09:02:19.520] [STDERR] Unhandled exception. System.ComponentModel.Win32Exception (8): Exec format error
[09:02:19.520] [STDERR]    at System.Diagnostics.Process.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
[09:02:19.520] [STDERR]    at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
[09:02:19.520] [STDERR]    at System.Diagnostics.Process.Start()
[09:02:19.520] [STDERR]    at Microsoft.Crank.Jobs.Bombardier.Program.Main(String[] args) in /private/var/folders/9n/ghtcrrtx7pn2jdd7fjxf33580000gn/T/benchmarks-agent/benchmarks-server-23139/tyoiw5zr.2iz/crank/src/Microsoft.Crank.Jobs.Bombardier/Program.cs:line 131
[09:02:19.520] [STDERR]    at Microsoft.Crank.Jobs.Bombardier.Program.<Main>(String[] args)
```

Any idea what's going wrong here? It almost looks like the bombardier executable cannot be run on my machine. I'm running on a Mac, so maybe that's the case. Is that not supported?

PR #83, opened 2020-07-31
Add support for running Bombardier job on Mac OS

jmezach:
After filing issue #81 I did some research to see if I can figure out why it wasn't working. Turns out it was indeed missing support for running this on Mac OS. Since I'm on a Mac I figured I might as well fix it myself.

I've rewritten the OS detection logic a little bit, because it seemed to be using some old API that doesn't accurately report the difference between Mac en Linux, which was needed in this case. Because of the security features in OSX I also had to run an additional command `spctl --add <executable>` to allow the downloaded executable to be run. Interestingly it asks me for permission to run the tool, but in the mean time it just runs it. But I guess that's an OSX issue.

### DIFF
--- a/src/Microsoft.Crank.Jobs.Bombardier/Program.cs
+++ b/src/Microsoft.Crank.Jobs.Bombardier/Program.cs
@@ -83,8 +83,8 @@ namespace Microsoft.Crank.Jobs.Bombardier
             using (var fileStream = File.Create(bombardierFileName))
             {
                 await downloadStream.CopyToAsync(fileStream);
-                if (Environment.OSVersion.Platform == PlatformID.Unix ||
-                    Environment.OSVersion.Platform == PlatformID.MacOSX)
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
                 {
                     Process.Start("chmod", "+x " + bombardierFileName);
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Crank.Jobs.Bombardier/Program.cs
```diff
@@ -62,17 +57,44 @@ static async Task Main(string[] args)
...
+            Console.WriteLine($"Downloading bombardier from {bombardierUrl} to {bombardierFileName}");
             using (var downloadStream = await _httpClient.GetStreamAsync(bombardierUrl))
             using (var fileStream = File.Create(bombardierFileName))
             {
                 await downloadStream.CopyToAsync(fileStream);
-                if (Environment.OSVersion.Platform == PlatformID.Unix)
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
                 {
+                    Console.WriteLine($"Setting execute permission on exeecutable {bombardierFileName}");
```
sebastienros, line 89:
```suggestion
                    Console.WriteLine($"Setting execute permission on executable {bombardierFileName}");
```

I would have changed it myself, by apparently you revoked the PR permissions (the default is to allow PRs to be edited by the owner of the targetted repo)

jmezach: Hmm, that's weird. I didn't get that question I think. Maybe that's because this is coming from a repo in another org? I'll check that. Suggested change looks good though, so I'll just commit it.

sebastienros:
Looks good ... but I don't understand why the CI doesn't run for this PR

jmezach:
Hmm, yeah, good question. I can see an azure-pipeline.yml in the repo, but I don't know why it's not triggering. Maybe the repo isn't linked to Azure DevOps somehow?

sebastienros:
It's working for my PRs, I am investigating.

sebastienros:
Found the problem. It's building now.

PR #96, opened 2020-08-11
Adding option to set agent build path

### LINKED ISSUE #82: Override temp folder
Can some configuration option for temporary folder where everything would be stored?

I would like to remove pressure from disk C if possible.

sebastienros:
Fixes #82


kant2002:
Thanks

PR #116, opened 2020-08-27
A sample application using a mock server

alefranz:
Hi @sebastienros 
I've created a sample application using wiremock as mock server.

Opening this mainly to discuss the approach and a few potential improvements I thought of with this experiment, before raising them as issues.

The idea of having a Wiremock job is that you can easily setup canned response in its json configuration, which is quite flexible. Not sure however if performance would be good enough (and how accurate is the applied delay).

Other comments in line

### DIFF
--- a/samples/client/Startup.cs
+++ b/samples/client/Startup.cs
@@ -3,10 +3,8 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
+using System.Net.Http;
 using System.Reflection;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
--- a/samples/client/Startup.cs
+++ b/samples/client/Startup.cs
@@ -1,0 +1,59 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace client
+{
+    public class Startup
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
+            {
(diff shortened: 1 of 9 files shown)


samples/client/mockserver/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM rodolpheche/wiremock as mockserver
+COPY config /home/wiremock
```
alefranz, line 2: I think it would be nice to add a 3rd way to add a docker job by specify the docker hub name, or did I miss it?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -1681,7 +1681,9 @@ private static async Task<(string containerId, string imageName, string workingD
 
             var command = OperatingSystem == OperatingSystem.Linux
                 ? $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --privileged --network host {imageName} {source.DockerCommand}"
-                : $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --network SELF --ip {hostname} {imageName} {source.DockerCommand}";
+                : string.Equals(hostname, "localhost", StringComparison.OrdinalIgnoreCase)
+                    ? $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} -p {job.Port}:{job.Port} {imageName} {source.DockerCommand}"
+                    : $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --network SELF --ip {hostname} {imageName} {source.DockerCommand}";
```
alefranz, line 1686: I had to run this to be able to run locally on windows without having to create a network and to be able to map the port.
Not sure however if that `Port` parameter is the correct parameter.

Maybe there should be a more generic argument to pass parameter to docker while I believe currently you can only specify arguemnts to the container

samples/client/mockserver/Dockerfile
alefranz: I think we could add another property `dockerSourceImage`, and when present instead of doing a `docker build` doing a `docker pull` then tagging the image with the `dockerImageName` so that the rest of the logic stay the same.
What do you think @sebastienros ? would you accept a PR for that?

src/Microsoft.Crank.Agent/Startup.cs
sebastienros: I think we can just keep the port:port version on Windows. I used SELF because I was trying to create a faster network interface, but at least if port:port works it's better for now.

samples/client/mockserver/Dockerfile
sebastienros: I agree with the suggestion. Had the same issue myself which required me to create a dumb docker file. `dockerImage` ? Would allow for name and tag. And change how dockerImageName is constructed when not provided to be sure the tag part is not interfering with the syntax of the command line.

samples/client/Startup.cs
```diff
@@ -0,0 +1,57 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
sebastienros, line 1: Could use the new WebApplicationBuilder model now

samples/client/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                endpoints.MapGet("/", async context =>
+                {
+                    var content = await client.GetStringAsync("/");
+
+                    await context.Response.WriteAsync(content);
+                });
+            });
+
+            Console.WriteLine($"AspNetCore location: {typeof(IWebHostBuilder).GetTypeInfo().Assembly.Location}");
```
sebastienros, line 49: Not necessary. Results show the version automatically.

sebastienros (commented):
I know ...

alefranz:
Hi @sebastienros !

I know .... 😂

I've not being using Crank recently. I wonder if this could still be useful or is there already an implemented alternative?

Thanks!

PR #129, opened 2020-09-17
Adding measurements rest endpoints

sebastienros:
/jobs/services/{serviceName}
/jobs/services/{serviceName}/measurements[?name={metric}]
/jobs/services/{serviceName}/measurements/last[?name={metric}]

Example:
http://localhost:5010/jobs/services/application/measurements/last?name=benchmarks/cpu


src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -720,5 +783,49 @@ public void Dispose()
             }
         }
 
+
+        private IActionResult SummaryResult(Func<Job> filter)
```
MarioszMSFT, line 787: this does not seem to be used.

MarioszMSFT (approved):

PR #138, opened 2020-09-29
Creating Event Sources package

sebastienros:
Creates a new NuGet source package to reference from benchmarked applications and clients in order to record measures.


davidfowl:
Why a source package?

sebastienros:
I was expecting this question. And apart from not having to publish another dll I don't remember. But I have that solution in mind for some time for a reason. Trying to figure out now ... If you think there are issues with it then I'll change it.

sebastienros:
Not using a source package anymore.

PR #147, opened 2020-10-16
BenchmarkDotNet support

sebastienros:
- [x] Documentation


sebastienros:
@roji With that we don't need custom parameters in the benchmark apps. See the example file that targets dotnet/performance micro benchmarks directly.

PR #155, opened 2020-10-26
Add additional averaging function (P95) for existing counters

MarioszMSFT:
Add additional averaging function (P95) for existing counters


sebastienros:
This is now supported oob

https://github.com/dotnet/crank/blob/master/docs/event_counters.md#building-results-out-of-measurements

PR #159, opened 2020-10-28
Optimizing polling apis

sebastienros:
- Skip MVC for most apis that are used to poll jobs, which could impact the current runs
- Removing locks and use ConcurrentDictionary


sebastienros:
on 100 RPS fixed rate

CPU:
Before (using /legacy route): 9.6% usage
After: 5.6%

PR #170, opened 2020-11-20
Fixing trace collection when jobs are stopped

sebastienros:
Collection was stopped when the application was stopped normally. The driver is expecting the application to stop in 30s and huge traces would put it in timeout.

The bug is that collection should not stop when Stopping is invoked in a normal path. The driver is supposed to call Stopping then TraceCollecting.


sebastienros:
Fixes #1602

Issue #160, opened 2020-10-28
Add script hook on results


sebastienros:
The goal is to be able to generate more results (and metadata?) from the measurements.
For instance compute the P95 of the GC metrics, which is based on an existing measure. Or render more than just one aggregating operation (max) without having to hard code all the potential variations.

This can be done by adding a new section that will contain some javascript to alter the results json payload. These scripts can be shared as they are part of the configuration files. The command line will then allow to run specific scripts to postprocess the data.

/cc @MarioszMSFT

Issue #100, opened 2020-08-11
CPU measurement precision


sebastienros:
Printing of benchmarks/cpu/raw prints far more digits of precision than the underlying system is capable of measuring.

Issue #10, opened 2020-04-27
Support the ability to collect custom event counters


davidfowl:
We should support specifying a set of event sources to collect counters from for a job.

PR #172, opened 2020-11-23
Adding scripts and custom providers

sebastienros:


src/Microsoft.Crank.Controller/README.md
```diff
@@ -114,7 +114,7 @@ Options:
     --application.endpoints http://asp-perf-lin:5001
     --application.sdkversion 5.0.100-alpha1-015721 
     --application.dotnetTrace true 
-    --application.collectCounters true
```
davidfowl, line 117: Thi ssitll works right?

sebastienros: No, it will only be by defining `--application.options.counterProviders System.Runtime`

I could keep the old way (too) for backward compatibility. Only issue is that it's not a server property like it used to be, so in theory it should go in "options". But it's only a logical constraints.

davidfowl: You should keep it for now.

sebastienros: What about I restore it, and add also `options.collectCounters` to provide the same behavior, and a warning to switch to the new one when it's used.

sebastienros:
Fixes #100
Fixes #160
Fixes #10

PR #175, opened 2020-11-27
dotnet load generators

sebastienros:


src/Microsoft.Crank.Jobs.HttpClient/httpclient.yml
```diff
@@ -0,0 +1,29 @@
...
+    html: '--header "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"'
+    json: '--header "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7"'
+    connectionclose: '--header "Connection: close"'
+  presetHeaders: none
+
+jobs:
+  httpclient:
+    source:
+      repository: https://github.com/dotnet/crank.git
+      branchOrCommit: sebros/httpclient
```
sebastienros, line 14: Local testing for now

src/Microsoft.Crank.Jobs.PipeliningClient/pipelining.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   pipelining:
     source:
       repository: https://github.com/dotnet/crank.git
-      branchOrCommit: master
+      branchOrCommit: sebros/httpclient
```
sebastienros, line 14: Local testing

davidfowl:
No review?

sebastienros:
@HaoK needs a client that we can add client certs to. So it might not be perfect as long as it works it will do.

PR #179, opened 2020-12-02
Display trace file download progress

sebastienros:
Just to make remote workers happy

image

@davidfowl @adamsitnik @roji


roji:
Yay!

(aren't all workers remote now?)

sebastienros:
There remote worker that work in more remote places, some on other continents, even islands.

PR #181, opened 2020-12-05
Migrate to diagnostics library

sebastienros:


sebastienros:
Forgot to migrate dotnet trace support ...

PR #187, opened 2020-12-09
Adding custom headers variable

sebastienros:
/cc @javiercn @HaoK


HaoK:
Nice!

sebastienros:
Reverting since I think it's causing a conflict. Will take more time to test it.

PR #197, opened 2020-12-16
Reuse source or builds between benchmarks

sebastienros:
Example: The application's source can be reused, so it's not cloned again. The load's build files can be reused, so it's not built again.

```
crank --config ..\..\samples\hello\hello.benchmarks.yml --scenario hello --profile local --application.options.reuseSource true --load.options.reuseBuild true
```

When a source parameter is changed, like a branch, or a project name, the source will still be cloned as the source key will change too. 
It doesn't detect the latest commit on a branch though. It could be done automatically by adding it to the computed SourceKey We could also add a custom "VaryBy" property to refresh the source for instance every day, or per user, ...

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -523,6 +523,31 @@ namespace Microsoft.Crank.Agent
                                 }
                                 else
                                 {
+                                    tempDir = null;
+
+                                    if (!String.IsNullOrEmpty(job.Source.SourceKey))
+                                    {
+                                        tempDir = Path.Combine(_rootTempDir, job.Source.SourceKey);
+
+                                        if (!Directory.Exists(tempDir))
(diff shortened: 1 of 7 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -523,6 +523,31 @@ private static async Task ProcessJobs(string hostname, string dockerHostname, Ca
...
                                 {
+                                    tempDir = null;
+
+                                    if (!String.IsNullOrEmpty(job.Source.SourceKey))
+                                    {
+                                        tempDir = Path.Combine(_rootTempDir, job.Source.SourceKey);
+
+                                        if (!Directory.Exists(tempDir))
+                                        {
+                                            Log.WriteLine("Creating resuable: " + tempDir);
```
MarioszMSFT, line 534: syntax

PR #201, opened 2020-12-19
Update performance issue title

sebastienros:


src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -271,7 +271,7 @@ private static async Task CreateRegressionIssue(IEnumerable<Regression> regressi
 
             var body = await CreateIssueBody(regressions, template);            
             
-            var title = "Performance regression: " + String.Join(", ", regressions.Select(x => x.CurrentResult.Scenario).Take(5));
+            var title = "Performance variation: " + String.Join(", ", regressions.Select(x => x.CurrentResult.Scenario).Take(5));
```
davidfowl, line 274: change?

sebastienros: I like it, also deviation 

Other options: difference · disparity · imbalance · alteration · modification · variability · variance · divergence · fluctuation · modification

davidfowl: difference

sebastienros:
This will change the title. I will change the summary of the issue in the templates next. These are in the aspnet/benchmarks repository where the configuration for our scenarios is kept.

PR #202, opened 2021-01-04
Add cert support to httpclient

HaoK:


src/Microsoft.Crank.Jobs.HttpClient/httpclient.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   httpclient:
     source:
       repository: https://github.com/dotnet/crank.git
-      branchOrCommit: master
```
HaoK, line 14: revert to master before merge

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -79,7 +87,33 @@ static async Task Main(string[] args)
                     }
                 }
 
-                return RunAsync();
+                CertPath = optionCertPath.Value();
```
sebastienros, line 90: `if (optionCertPath.HasValue())`

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -79,7 +87,33 @@ static async Task Main(string[] args)
...
+                Console.WriteLine("CerPath: " + CertPath);
+                if (CertPath != null && !string.Equals(CertPath, "none", StringComparison.OrdinalIgnoreCase))
```
sebastienros, line 92: "none" can be removed I think, even more if `optionCertPath.HasValue()` is used

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -261,8 +311,9 @@ public static async Task<WorkerResult> DoWorkAsync()
                         counters[status / 100 - 1]++;
                     }
                 }
-                catch
+                catch (Exception e)
                 {
+                    Console.WriteLine(e);
```
sebastienros, line 316: Remove now to prevent flooding the log when it goes wrong? Or maybe only on `socketErrors == 0`?

HaoK:
Updated @sebastienros

PR #218, opened 2021-02-01
Fix documentation for collecting event counters

mgasparel:
I had some issues following the docs for collecting System.Runtime counters. Specifically, using `--application.counterProviders System.Runtime` as recommended resulted in the error `Could not find part of the configuration path: '[application.counterProviders, System.Runtime]'`

After digging around a bit, I noticed this should actually be `application.options.counterProviders`

This PR is just a quick fix for the documentation.


sebastienros:
Thanks, forgot to update the doc when I moved it.

PR #227, opened 2021-02-09
Adding doc to setup the agent on linux

sebastienros:


docs/setup_linux.md
```diff
@@ -0,0 +1,74 @@
...
+
+To display the live log, run the following command:
+
+```
+docker logs -f --tail 100 crank-agent
+```
+
+### Continuous integration
+
+In order to restart and update the agent regularly, the follow cron job can be used.
```
RobertHenry6bev, line 60: s/follow/following/

docs/setup_linux.md
```diff
@@ -0,0 +1,74 @@
+## Setting up an agent on Linux
+
+On Linux, it is recommended to setup the Agent using the Docker file provided in this repository.
+There instructions are valid for x86 and arm.
```
RobertHenry6bev, line 4: s/x86/x86_64/
s/arm/ARM64 (aka ARMv8 or AARCH64)/

sebastienros: Where is this feedback syntax coming from? I get it, took me the second example though.

PR #244, opened 2021-02-26
Turn on preemptive for all mono scenarios

fanyang-mono:


fanyang-mono:
Will do this in yml file instead.

PR #267, opened 2021-03-16
Modernize the Microsoft.Crank.AzureDevOpsWorker project

IEvangelist:
Use C# 9

- Encapsulation:
  - Common "Send" async `bool` returning methods, see:
   
```csharp
private async Task<bool> SendAsync<TBody>(
    string url, TBody body, [CallerMemberName] string callerName = "")
{
    var requestBody = JsonSerializer.Serialize(body);

    try
    {
        var result = await PostDataAsync(url, requestBody);
        return result.IsSuccessStatusCode;
    }
    catch (Exception ex)
    {
        Console.WriteLine($"{callerName} failed: {url}");
        Console.WriteLine(ex);
        return false;
    }
}
```

Add [caller member name](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-5.0), to encapsulate code (a bit more DRY)

- [Use `string` instead of `String` 🤷🏽‍♂️](https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html)
- Use "Async" suffix
- Use string interpolation
- Delete un-used class `ProcressResult`
- Fix bug where `Job.FlushStandardError()` incorrectly returned stdout instead
- Add `<Nullable>enable</Nullable>` to the project.
- Use pattern matching where it makes sense
- Use static local functions when preferred
- Use expression bodied members / .ctor
- Use range operator instead of `string.Substring()`
- Be a bit more defensive in our coding
  - `sealed` all the things, removed a warning for our implementations of `IDisposable` as well
  - Mark certain fields as `readonly`
  - Reduce the scope `var jobPayload`, demote from method-level, to sub-routine
- Change `Program.Main` to `Task<int>` instead of `int`
  - `using` app and `return app.ExecuteAsync(args);`
- Use target-type new, where and when obvious
- Add `#pragma CS8618` since we'd rather not register event handlers in the `.ctor`, but due to `nullable` two events aren't assigned, and we're warned about it
- Ensure that implementations of `IDisposable` are properly disposed of
  - Lift the scope of a few method variables to class scope, this was intentional to capture them such that they could have `.Dispose()` called on them.
  - Corresponding parent objects now themselves implement `IDisposable`, thus being cascaded-friendly.
  - Objects are now in `using` where appropriate


src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -191,49 +157,41 @@ public async Task<bool> UpdateTaskTimelineRecordAsync(string taskLogObject)
...
-        {
-            var buffer = Encoding.UTF8.GetBytes(requestBody);
-            var byteContent = new ByteArrayContent(buffer);
-            byteContent.Headers.ContentType = new MediaTypeHeaderValue("application/json");
-
-            var response = await _httpClient.PatchAsync(new Uri(url), byteContent);
-
-            return response;
-        }
+        public void Dispose() => _httpClient?.Dispose();
```
sebastienros, line 195: The `_httpClient` instance is static, it's reused across messages during the whole app lifetime. This can't be disposed.

src/Microsoft.Crank.AzureDevOpsWorker/Job.cs
```diff
@@ -150,29 +134,27 @@ public void Stop()
...
-            while (_standardOutput.TryDequeue(out var result))
+            while (queue is { }
+                && queue.TryDequeue(out var result))
             {
                 yield return result;
             }
         }
 
-        public bool WasSuccessful => _process != null && _process.ExitCode == 0;
+        public bool WasSuccessful => _process is { ExitCode: 0 };
```
sebastienros, line 155: Learning about this syntax ;)

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
IEvangelist: That is a good callout. I only saw it being used `Program` but see it's called multiple times on the event callback - missed that.

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -5,16 +5,17 @@
...
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using Azure.Messaging.ServiceBus;
 
 namespace Microsoft.Crank.AzureDevOpsWorker
 {
-    public class DevopsMessage
+    public sealed class DevopsMessage : IDisposable
```
IEvangelist, line 16:
```suggestion
    public sealed class DevopsMessage
```

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
IEvangelist, line 195:
```suggestion
```

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -163,38 +152,39 @@ private static async Task MessageHandler(ProcessMessageEventArgs args)
...
-                catch (Exception f)
+                catch (Exception ex)
                 {
-                    Console.WriteLine("Failed to abandon the message: " + f.ToString());
+                    Console.WriteLine($"Failed to abandon the message: {ex}");
                 }
             }
             finally
             {
+                devopsMessage?.Dispose();
```
IEvangelist, line 180:
```suggestion
```

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -38,13 +39,15 @@ public DevopsMessage(ServiceBusReceivedMessage message)
             TaskInstanceId = (string)message.ApplicationProperties["TaskInstanceId"];
             AuthToken = (string)message.ApplicationProperties["AuthToken"];
 
-            _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic",
-                    Convert.ToBase64String(Encoding.ASCII.GetBytes(":" + AuthToken)));
+            _httpClient.DefaultRequestHeaders.Authorization =
+                new AuthenticationHeaderValue("Basic",
```
davidfowl, line 43: You can target type here as well no?

IEvangelist: Yes, I believe so. I'll have to rebase this PR and address the merge conflicts, not sure how I missed the reply here. 😬

IEvangelist:
Closing at this continues to drift and I do not have the proper amount of time to try to keep up. 😬

PR #272, opened 2021-03-20
Fix multi-client metadata

### LINKED ISSUE #260: Multi endpoints broken
Can be tested with a single agent

### LINKED ISSUE #273: Latencies use wrong reduce function
Currently sum. Should be max, or avg.

sebastienros:
Fixes #260
Fixes #273

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -115,12 +115,13 @@ namespace Microsoft.Crank.Agent
         public static OperatingSystem OperatingSystem { get; }
         public static Hardware Hardware { get; private set; }
         public static string HardwareVersion { get; private set; }
-        public static Dictionary<Database, string> ConnectionStrings = new Dictionary<Database, string>();
         public static TimeSpan DriverTimeout = TimeSpan.FromSeconds(10);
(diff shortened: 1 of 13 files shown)


davidfowl:
No test

sebastienros:
> No test 

Need to fix multi-client on a single agent first. But this is unblocking the multi-agents scenario.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4961,6 +4941,15 @@ private void OnShutdown()
         {
             try
             {
+                Log.WriteLine("Cancelling remaining jobs");
+
+                if (!_processJobsCts.IsCancellationRequested)
+                {
+                    _processJobsCts.Cancel();
+
+                    Task.WaitAny(_processJobsTask, Task.Delay(TimeSpan.FromSeconds(30)));
```
davidfowl, line 4950: Why is it blocking here?

sebastienros: I want to let the currently running jobs, if any, to be stopped, or forced to stop, such that there is no ghost process holding on http ports. Without that the unit tests would fail, showing the issue.

sebastienros: Or is the question about "here"?

PR #285, opened 2021-04-10
Split metrics with multiple keys

sebastienros:


sebastienros:
Goal is to migrate from keys like `wrk/latency/mean` and `bombardier/latency/mean` to `http/latency/mean` such that it's easier to switch clients with tooling.

Plan:

1- Deploy new crank controller that supports multiple keys in results
2- Deploy clients updates that create results with multiple keys
3- Update charts to use the new metrics
4- Migrate old benchmarks results and stop using old metrics

PR #288, opened 2021-04-12
Pin extensions packages version

sebastienros:


davidfowl:
I don't like this PR. It's too hacky for the main crank repo. Can we make something and put this hardcoded list in aspnet/benchmarks?

davidfowl (approved):

PR #290, opened 2021-04-20
Improve runtimes version verification

sebastienros:


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3007,6 +2981,38 @@ long DirSize(DirectoryInfo d)
...
+                    return "win-arm64";
+                }
+                else
+                {
+                    return "win-x64";
+                }
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return "osx-x64";
```
RobertHenry6bev, line 2999: M1?

sebastienros: Not specifically, just moved some code around as I needed to reuse it, I don't know when it got in.
Feel free to buy one and try though ;)

PR #293, opened 2021-04-23
Call Stop on tracing EventPipeSessions

sywhang:
`EventPipeSession.Stop` needs to be called for the tracing session to successfully receive rundown events. 

In addition, counter sessions don't need rundown so we should specify requestRundown to false while starting the counter session to avoid unnecessary cost of emitting rundown.


sebastienros:
Thanks. Feel free to give some feedback on the rest of the code.

davidfowl:
Thanks @sywhang!

PR #297, opened 2021-04-28
Saving dependencies information

### LINKED ISSUE #295: Track versions and source information for every dependency that is part of a benchmark run
Make it possible to store version information and source information of every dependency when running a benchmark. This is important for tools to track what changes might have improved or regressed benchmarks. Today we have the ability to track the .NET SDK version, the ASP.NET Core version and the .NET Runtime version. We need the ability to track any package version or dll that was part of the run as well. This will give us more precise information and the ability to detect changes in more things when looking for regressions.

### LINKED ISSUE #296: Track version/sha information of the benchmark project
Today it's hard to know if changes to the benchmark itself cause an improvement or regression because we don't store this information run to run. This should be stored along with [dependency information](https://github.com/dotnet/crank/issues/295) in order to make it easier to root cause the source of change performance.

[This change](https://github.com/aspnet/Benchmarks/pull/1640) to the ASP.NET Core benchmarks repo caused a huge regression in one of the benchmarks https://github.com/dotnet/aspnetcore/issues/30713 and it took a while to figure out why. Having the benchmark information would have made it more obvious what change caused this.

sebastienros:
Fixes #295
Fixes #296

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -37,6 +37,7 @@ using NuGet.Versioning;
 using System.Net;
 using System.Reflection.PortableExecutable;
 using System.Reflection.Metadata;
+using System.Security.Cryptography;
 
 namespace Microsoft.Crank.Agent
 {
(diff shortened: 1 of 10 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2998,12 +2999,30 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
...
 
+            void CreateDependenciesHash()
+            {
+                using (var md5 = MD5.Create())
+                {
+                    foreach (var dependency in job.Dependencies)
+                    {
+                        var names = String.Concat(dependency.Names);
+                        byte[] bytes = md5.ComputeHash(Encoding.UTF8.GetBytes(names));
+                        dependency.Id = Convert.ToBase64String(bytes);
```
davidfowl, line 3021: Convert to hex? I guess it doesn't matter.

PR #310, opened 2021-05-13
Add container size metrics

billwert:
This adds both uncompressed and compressed (for wire size) metrics.

New output from application build:
| application                |         |
| -------------------------- | ------- |
| CPU Usage (%)              | 0       |
| Cores usage (%)            | 0       |
| Working Set (MB)           | 16      |
| Build Time (ms)            | 13,581  |
| Start Time (ms)            | 422     |
| Published Size (KB)        | 238,996 |
| Compressed Image Size (KB) | 94,880  |

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -639,19 +639,6 @@ namespace Microsoft.Crank.Agent
                                         });
                                     }
 
-                                    if (!job.Metadata.Any(x => x.Name == "benchmarks/compressed-image-size"))
-                                    {
-                                        job.Metadata.Enqueue(new MeasurementMetadata
-                                        {
-                                            Source = "Host Process",
-                                            Name = "benchmarks/compressed-image-size",
-                                            Aggregate = Operation.Max,
-                                            Reduce = Operation.Max,
-                                            Format = "n0",
-                                            LongDescription = "The size of the compressed docker image (KB)",
-                                            ShortDescription = "Compressed Image Size (KB)"
-                                        });
-                                    }
 
                                     if (!job.Metadata.Any(x => x.Name == "benchmarks/memory/swap"))
                                     {
@@ -1906,34 +1893,7 @@ namespace Microsoft.Crank.Agent
                         }
                     }
 
-                    var dockerSaveArguments = $"save {imageName} -o {imageName}.tar";
-
-                    var saveResults = await ProcessUtil.RunAsync("docker", dockerSaveArguments,
-                        workingDirectory: srcDir,
-                        cancellationToken: cancellationToken,
-                        log: true,
-                        outputDataReceived: text => job.BuildLog.AddLine(text));
-                    
-                    var gZipArguments = $"{imageName}.tar";
-
-                    var gZipResults = await ProcessUtil.RunAsync("gzip", gZipArguments,
-                        workingDirectory: srcDir,
-                        cancellationToken: cancellationToken,
-                        log: true,
-                        outputDataReceived: text => job.BuildLog.AddLine(text));
-                    
-                    var filePath = Path.Combine(srcDir, $"{imageName}.tar.gz");
-                    var compressedSize = new FileInfo(filePath).Length;
-                    if (compressedSize != 0)
-                    {
-                        job.Measurements.Enqueue(new Measurement
-                        {
-                            Name = "benchmarks/compressed-image-size",
-                            Timestamp = DateTime.UtcNow,
-                            Value = compressedSize / 1024
-                        });
-                        File.Delete(filePath);
-                    }
+                  
                 }
                 else
                 {
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -1895,13 +1895,13 @@ namespace Microsoft.Crank.Agent
                     {
                         if (imageSize != 0)
                         {
-                            job.PublishedSize = imageSize;
+                            job.PublishedSize = imageSize / 1024;
 
                             job.Measurements.Enqueue(new Measurement
                             {
                                 Name = "benchmarks/published-size",
                                 Timestamp = DateTime.UtcNow,
-                                Value = imageSize
+                                Value = imageSize / 1024
                             });
                         }
                     }
@@ -1930,7 +1930,7 @@ namespace Microsoft.Crank.Agent
                         {
                             Name = "benchmarks/compressed-image-size",
                             Timestamp = DateTime.UtcNow,
-                            Value = compressedSize
+                            Value = compressedSize / 1024
                         });
                         File.Delete(filePath);
                     }
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -1867,6 +1881,61 @@ private static async Task<(string containerId, string imageName, string workingD
...
+                            job.Measurements.Enqueue(new Measurement
+                            {
+                                Name = "benchmarks/published-size",
+                                Timestamp = DateTime.UtcNow,
+                                Value = imageSize
+                            });
+                        }
+                    }
+
+                    var dockerSaveArguments = $"save {imageName} -o {imageName}.tar";
```
sebastienros, line 1909: Why should we store the compressed size? Tracking it won't be different than tracking the non-compressed one? Any regression in one will be seen in the other. I assume "we" already know what is the size of the dotnet images already. If it was free that would not matter much, but here we zip the 200MB for no value (IMO, until convinced otherwise).

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -1867,6 +1881,61 @@ private static async Task<(string containerId, string imageName, string workingD
                     {
                         job.Error = job.BuildLog.ToString();
                     }
+
+                    var dockerInspectArguments = $"inspect -f \"{{{{ .Size }}}}\" {imageName}";
```
sebastienros, line 1885: Does it actually work? At first I was wondering if the resulting unit was actually KB, but then it appeared it didn't work for me since it's an image, not a container, so you can't do inspect. What worked was `docker images golang --format "{{ .Size }}"` and the result contains the unit, like `862MB`

sebastienros: For a container, the property was `{{ .SizeRw }}` in bytes

src/Microsoft.Crank.Agent/Startup.cs
billwert: I think understanding the impact to compressed size of various aot/linker options is interesing. @mangod9 what do you think?

src/Microsoft.Crank.Agent/Startup.cs
billwert: Yes, it works. On my machine it produces the same output as `docker image inspect`. I think Docker is smart enough to figure out what you're asking for, so passing it an image name does the correct thing.

The raw value looks like this:
```
        "Size": 244714070,
```
Which lines up with the size from `docker image ls`:
```
benchmarks  latest   275ac07cf3fb   2 weeks ago     245MB
```

However, you're right that I'm actually logging bytes, not KB. I'll fix that.

full output:
<details>

```
❯ docker inspect  benchmarks
[
    {
        "Id": "sha256:275ac07cf3fb6d34d33dc6b72f21458d951fbffdb7e05a3f699835c8e17055f4",
        "RepoTags": [
            "benchmarks:latest"
        ],
        "RepoDigests": [],
        "Parent": "",
        "Comment": "buildkit.dockerfile.v0",
        "Created": "2021-04-24T20:19:10.5097035Z",
        "Container": "",
        "ContainerConfig": {
            "Hostname": "",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": null,
            "Cmd": null,
            "Image": "",
            "Volumes": null,
            "WorkingDir": "",
            "Entrypoint": null,
            "OnBuild": null,
            "Labels": null
        },
        "DockerVersion": "",
        "Author": "",
        "Config": {
            "Hostname": "",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                "ASPNETCORE_URLS=http://+:80",
                "DOTNET_RUNNING_IN_CONTAINER=true",
                "DOTNET_VERSION=6.0.0-preview.4.21221.4",
                "ASPNET_VERSION=6.0.0-preview.4.21221.10",
                "Logging__Console__FormatterName=Json"
            ],
            "Cmd": null,
            "Image": "",
            "Volumes": null,
            "WorkingDir": "/app",
            "Entrypoint": [
                "dotnet",
                "Benchmarks.dll"
            ],
            "OnBuild": null,
            "Labels": null
        },
        "Architecture": "amd64",
        "Os": "linux",
        "Size": 244714070,
        "VirtualSize": 244714070,
        "GraphDriver": {
            "Data": {
                "LowerDir": "/var/lib/docker/overlay2/qn271qirrqnz8ydzs72xzjoxx/diff:/var/lib/docker/overlay2/4f5a007e0140f2504866786c255afed9aec0f53e36d91225c838f5e4b6d0c1dc/diff:/var/lib/docker/overlay2/09160f6171461ccd54515f8001613a31e6e53f4e6908460413d2fd336d34670e/diff:/var/lib/docker/overlay2/afdadcc2986c542ee596d4309bf197c3589210d6a0d62327ee19fd385d2a6911/diff:/var/lib/docker/overlay2/65244f19344582ebcd707a4b1d00ec2f91ad8550026b9a66bb0918c647079d16/diff:/var/lib/docker/overlay2/3b253c4070a8ee589d32d942ef68a59ace4ed612d6cf3a8d5b110b3b1bc5d824/diff",
                "MergedDir": "/var/lib/docker/overlay2/xdcfdtkea3b4uja9c86nwzs5c/merged",
                "UpperDir": "/var/lib/docker/overlay2/xdcfdtkea3b4uja9c86nwzs5c/diff",
                "WorkDir": "/var/lib/docker/overlay2/xdcfdtkea3b4uja9c86nwzs5c/work"
            },
            "Name": "overlay2"
        },
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:f353f969c442dec0cd79233d0c0db0ac1458f3811f118587eae5101fb50408a2",
                "sha256:3e08f253054617fd343125239d548ccc7aa33cb3b2296c65ea052afe24dc3452",
                "sha256:683948709559072f615a316b336f2d94f0140a3f5f69eb27c72545cc1321d4aa",
                "sha256:77f03ba195a91b8bf2d444e22bad0837ff955b0a3194dfb85dda18a82fb04ade",
                "sha256:1c07b3a04d96986b4d863a7673e027aafb6ebef2bf4d45498b00ed1be47f4995",
                "sha256:94a95068aa6d3aa9cc9c19edc3b29b1258aa69a57f4f9a5953439dfb441703b5",
                "sha256:bb4b627338bbab2710a680c81daaf67165ae9d14d6bcc66319cb66df6441953a"
            ]
        },
        "Metadata": {
            "LastTagTime": "2021-04-24T20:19:10.8237854Z"
        }
    }
]
```

</details>

src/Microsoft.Crank.Agent/Startup.cs
billwert: Also @richlander

mangod9: is the concern that we a single metric is too costly to store? Ok with not storing at the moment.

billwert: I think the concern is the time to collect it. We spend an extra 10 seconds or so to create the zip file.

@sebastienros its gonna be the same file for every run in this mix. What if I make it a conditional based on some property like I originally suggested and just do it for a single test? That way the whole matrix doesn't spend this time.

sebastienros: It has a cost of time to serialize the image to a file, then run the command to compress it, for each run of the matrix, just to know what is the size of the container once compressed. I don't see the value of computing it on each run. It should be a known value. Perf changes won't impact the result run after run, and raw publish size changes will show us any regression in size.

sebastienros: Still seems odd. It shouldn't be the agent's concern to measure such an artificial value. I am not sure anyone has told me yet what value it provides to have it there.

Please educate me. Even outside of the agent, do we want to know what is the compressed size of the final image with the app? If so do we know it already? Either we don't know in which case that can't be that important and we can do a single run to know, or we do and that won't change much more than the uncompressed size.

An argument I could accept is that we want to compare the same app to a different framework and see how big the image is. But how much value does it add to comparing the images sizes in their uncompressed state. And how much flexibility do we have to improve the compressed size "more" than the uncompressed size?

The code is there, and we can just merge the PR as-is, but my argument is more articulated reasons to measure something, and the value it adds. 

@billwert I am ok with adding an argument in the Job (`bool MeasureCompressedDockerSize`, undocumented), running it for all the scenarios (easier to chart than a single one), and in a few weeks we look at what knowledge all these daily measurements got us, and if there is any reason to keep this running.

billwert: It would have been handy to have had it for blogging purposes. That said, I don't feel too strongly about it and I've just removed the code.

Can we merge?

PR #329, opened 2021-09-10
[Doc] Fixed the crank command in "Testing local changes"

corrieriluca:
Hi,

I noticed a small mistake while reading the "Benchmarking local applications" documentation.

The change made in the `Main` method of the program is not taken into account with the given command since the YAML benchmark file used is the one which fetches source from a GitHub repo. It should be the YAML file already used throughout the page (i.e. `sample/local/local.benchmarks.yml`).

Regards

### DIFF
--- a/docs/local_application.md
+++ b/docs/local_application.md
@@ -61,7 +61,7 @@ public static void Main(string[] args)
 Execute the following command:
 
 ```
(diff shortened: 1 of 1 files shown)


sebastienros:
Thanks

PR #330, opened 2021-09-10
[Doc/Samples] Rename variables in microbenchmarks YAML samples to avoid conflict

corrieriluca:
Hi,

I stumbled upon an issue while reading and testing the "Running micro-benchmarks" section of the documentation.

In the example below (from the `micro.benchmarks.yml` sample), the `job` variable does not expand as expected (we expect `short`) at runtime, and thus results in a error from BenchmarkDotNet (it fails parsing its arguments).

```yaml
  benchmarks:
    source:
      localFolder: .
      project: micro.csproj
    variables:
      filter: "*"
      job: short
    arguments: --job {{job}} --filter {{filter}} --memory
    options:
      benchmarkDotNet: true
```

Here is the output from `crank-agent` (shrinked) : 

```
[11:07:27.276]   Arguments:  --inProcess --cli C:\Users\[...]\src\published\micro.exe --join --exporters briefjson markdown --job {
  "DriverVersion": 0,
  "ServerVersion": 4,
  "Id": 0,

  [...]

  "Dependencies": [],
  "CollectDependencies": false
} --filter * --memory
```

It seems (according to the [`benchmarks.schema.json`](https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Controller/benchmarks.schema.json) file) that `job` is already used as a global variable by `crank` to define the job to be executed. That explains why the `job` variable expands to a big JSON block in the output above.

In this PR, I suggest renaming these variables in the `micro.benchmarks.yml` and `dotnet.benchmarks.yml` to `filterArg` and `jobArg` in order to correct this issue and avoid any conflict.

Eventually, I replaced the reference to `master` by `main` in `dotnet.benchmarks.yml` for the `dotnet/performance` repo.

Regards,

### DIFF
--- a/docs/microbenchmarks.md
+++ b/docs/microbenchmarks.md
@@ -54,9 +54,9 @@ The following example is available at https://github.com/dotnet/crank/blob/main/
       localFolder: .
       project: micro.csproj
     variables:
-      filter: "*"
-      job: short
-    arguments: --job {{job}} --filter {{filter}} --memory
+      filterArg: "*"
+      jobArg: short
+    arguments: --job {{jobArg}} --filter {{filterArg}} --memory
     options:
       benchmarkDotNet: true
 ```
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/crank?pullRequest=330) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/crank?pullRequest=330) before we can accept your contribution.<br/><br/>:x: corrieriluca [sign now](https://cla.dotnetfoundation.org/dotnet/crank?pullRequest=330)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/crank?pullRequest=330) it.</sub>

sebastienros:
Thanks. I indeed introduced the `job` global property recently.

PR #365, opened 2021-11-26
Adds support for running CrankAgent as a windows service

rducom:
We can't run Crank.Agent as a windows service.
This PR adds an opt-in parameter to let Crank.Agent be windows-service compatible.

### DIFF
--- a/src/Microsoft.Crank.Agent/Microsoft.Crank.Agent.csproj
+++ b/src/Microsoft.Crank.Agent/Microsoft.Crank.Agent.csproj
@@ -19,6 +19,7 @@
 
   <ItemGroup>
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="3.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.WindowsServices" Version="5.0.12" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.8" />
     <PackageReference Include="Microsoft.Azure.Relay.AspNetCore" Version="1.3.15173" />
     <PackageReference Include="Microsoft.Diagnostics.NETCore.Client" Version="0.2.251802" />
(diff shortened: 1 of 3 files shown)


sebastienros:
Awesome! Does it mean that if it's started once, then it will be run automatically on next machine reboot? Have you tried it with your changes? I could definitely use it. Right now I run a script on startup to pull the latest source code, build, clean NuGet cache and start it. What about logging, how should we check what's happening? Might have to switch to ILogger instead of Console.WriteLine calls.

rducom:
> Does it mean that if it's started once, then it will be run automatically on next machine reboot?

It means you can setup a windows service with `SC CREATE "CrankAgentService" binpath= "C:\....\crankagent.exe parameter-list'`, and configure it to automatically start (and be up on next reboot). We are currently building our crank infrastructure, and getting the Agent running as a windows service with this PR. 

I can also add support for systemD if needed (we don't have a linux VM crank agent yet, it's a matter of days/weeks)

> What about logging

Good point! Didn't thought about it. It's obvious we need logs from the agent!
Crank.Agent loglevel is currently using `.SetMinimumLevel(LogLevel.Error);`. 
Are you OK if I set it at `LogLevel.Information`, and :
- replace all Console.WriteXXX with logger.LogInformation(...)
- adds an optional parameter for a custom logPath

sebastienros:
> adds an optional parameter for a custom logPath

Shouldn't it be part of the logger configuration? So using a custom logger (NLog?) as the default ASP.NET one doesn't support files (I don't thing it has been done in 6.0 even).

OK for everything else.

rducom:
> Shouldn't it be part of the logger configuration?

For users running directly from the source code, or using "published assemblies", an appsettings.json will be convenient, but I'm also concerned for users running Microsoft.Crank.Agent as a dotnet tool : how can they control/edit such an appsettings.json ? I'm not aware of appsettings patterns for dotnet tools.

I was thinking about using Serilog, but I'm open to any third party log-provider. No pb if you prefer NLog.

sebastienros:
How do other services handle logging? I assume a single argument might not be sufficient in some cases, you don't want the same file to grow indefinitely, at least a rolling file would be better.

rducom:
> How do other services handle logging? 

I believe the "windows standard" is using EventLog sink (output is the Windows Event Log). 
But usability is low, and log volume for any AspNetApp (like Crank.Agent) will be high...

> you don't want the same file to grow indefinitely

Yes, rolling log files are mandatory.

sebastienros:
Can you add a minimal usage example in the README of the agent? I would merge this PR even without other features for now since it can unblock you.

PR #369, opened 2021-12-02
Adds Crank.Agent log file support - serilog

rducom:
As discussed in https://github.com/dotnet/crank/pull/365, this PR adds log file support for crank agent.
Users can pass an opt-in `--log-path` command-line argument with a full path to a log-file.

Log files roll every 1 day.

### DIFF
--- a/src/Microsoft.Crank.Agent/Microsoft.Crank.Agent.csproj
+++ b/src/Microsoft.Crank.Agent/Microsoft.Crank.Agent.csproj
@@ -29,6 +29,11 @@
     <PackageReference Include="NuGet.Versioning" Version="5.5.1" />
     <PackageReference Include="Vanara.PInvoke.Kernel32" Version="3.3.14" />
     
+    <PackageReference Include="Serilog.Extensions.Logging" Version="3.0.1" />
+    <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="4.0.0" />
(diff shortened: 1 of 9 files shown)


rducom:
@sebastienros I have an build error related to usage of Serilog

`error SIGN001: Signing 3rd party library Serilog.dll with Microsoft certificate 'Microsoft400'. The library is considered 3rd party library due to its copyright: ''`

I wasn't expecting it, since [Serilog is used](https://github.com/microsoft/reverse-proxy/search?q=serilog) in Yarp without issues. What is Crank policy about third party lib usage ?

sebastienros:
Needs to be registered here: https://github.com/dotnet/crank/blob/main/eng/Signing.props
And also added to https://github.com/dotnet/crank/blob/main/NOTICE.txt

src/Microsoft.Crank.Agent/Microsoft.Crank.Agent.csproj
```diff
@@ -29,6 +29,11 @@
     <PackageReference Include="NuGet.Versioning" Version="5.5.1" />
     <PackageReference Include="Vanara.PInvoke.Kernel32" Version="3.3.14" />
     
+    <PackageReference Include="Serilog.Extensions.Logging" Version="3.0.1" />
```
sebastienros, line 32: That's a lot of direct references just for a file logger. Are they all required?

rducom: You are right, Serilog.AspNetCore includes Serilog.Extensions.Logging, so not needed

PR #375, opened 2021-12-16
elasticsearch indexing of result

abimbolaomotoso:


src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -1102,6 +1134,15 @@ string EscapeCsvValue(string value)
                     }
                 }
 
+ if (!String.IsNullOrEmpty(_elasticSearchUrl))
```
sebastienros, line 1137: Formatting is wrong across the document

src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -36,7 +36,8 @@ public class Program
 
         private static string _tableName = "Benchmarks";
         private static string _sqlConnectionString = "";
-
+ private static string _indexName = "Benchmarks";
```
sebastienros, line 39: If it's case sensitive then maybe use lowercase?

src/Microsoft.Crank.Controller/Program.cs
abimbolaomotoso: oh! I'll attend to it and get back

abimbolaomotoso: Fixed

src/Microsoft.Crank.Controller/Program.cs
abimbolaomotoso: Fixed

abimbolaomotoso:
Hello @sebastienros,
I'm I missing anything? Kindly let me know so I can take care of it.
Thanks

sebastienros (approved):

PR #391, opened 2022-02-03
Add net6.0 TFM to tools

martincostello:
Add a `net6.0` TFM for the .NET tools.

I've been playing with crank in GitHub Actions today with a repo that only uses .NET 6, and as .NET 5 wasn't installed as well, the tools would fail to run.

You can work around this by setting `DOTNET_ROLL_FORWARD=Major` so you don't need to install the .NET 5 runtime as well, but adding the additional TFM should make them **Just Work™️** without the need to do that either.


sebastienros:
I didn't think you could multi-target a dotnet tool 🤦

sebastienros (approved):

PR #397, opened 2022-02-04
Add job options to download output and build logs

benjaminpetit:
Not a big fan of `DownloadOutputOutput` but I tried to keep naming consistent

### DIFF
--- a/src/Microsoft.Crank.Controller/JobConnection.cs
+++ b/src/Microsoft.Crank.Controller/JobConnection.cs
@@ -1013,11 +1013,32 @@ namespace Microsoft.Crank.Controller
             }
         }
 
-        public async Task<string> DownloadBuildLog()
+        public async Task DownloadBuildLogAsync()
         {
-            var uri = Combine(_serverJobUri, "/buildlog");
+            if (!Job.Options.DownloadBuildLog)
+            {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Crank.Controller/JobConnection.cs
```diff
@@ -1013,11 +1013,32 @@ private static void CleanTemporaryFiles()
...
-        public async Task<string> DownloadBuildLog()
+        public async Task DownloadBuildLogAsync()
         {
-            var uri = Combine(_serverJobUri, "/buildlog");
+            if (!Job.Options.DownloadBuildLog)
+            {
+                return;
+            }
+
+            var fileName = Path.Combine(Job.Options.DownloadFilesOutput ?? ".", $"{Job.RunId}.{_jobName}.{_serverUri.Host}.{Job.Id}.build.log");
```
sebastienros, line 1023: I think this should handle filenames passed directly. And otherwise use the job name and datetime. This logic is already used in other locations that download files, like traces.

sebastienros:
It's also missing the updated README with the new arguments. I think there is also a static string containing the JOB specific options documentation.

sebastienros (approved):

PR #406, opened 2022-02-08
Fix perfcollect and nghttp2 installation

sebastienros:


docker/agent/Dockerfile
```diff
@@ -47,14 +47,15 @@ RUN apt-get update \
...
+# 
+ENV NGHTTP2_VERSION 1.46.0
 RUN cd /tmp \
     && curl -L "https://github.com/nghttp2/nghttp2/releases/download/v${NGHTTP2_VERSION}/nghttp2-${NGHTTP2_VERSION}.tar.gz" -o "nghttp2-${NGHTTP2_VERSION}.tar.gz" \
     && tar -zxvf "nghttp2-${NGHTTP2_VERSION}.tar.gz" \
     && cd /tmp/nghttp2-$NGHTTP2_VERSION \
     && ./configure \
     && make \
-    && make install
+    && make install || true
```
BrennanConroy, line 58: Seems odd, wouldn't we want to fail if install doesn't work?

sebastienros: It's failing on a single machine, and I can't explain why. And we don't use this machine to run http2 benchmarks. So for now I am ignoring it.

BrennanConroy (approved):

PR #416, opened 2022-03-04
React to new SDK version format

sebastienros:
File format has changed for https://aka.ms/dotnet/7.0.1xx/daily/productCommit-win-x64.txt

https://github.com/dotnet/installer/pull/13132

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -3908,7 +3908,7 @@ namespace Microsoft.Crank.Agent
             }
             else if (String.Equals(sdkVersion, "Edge", StringComparison.OrdinalIgnoreCase))
             {
-                (sdkVersion, _) = await ParseLatestVersionFile(_latestSdkVersionUrl);
+                (sdkVersion, _) = await ParseVersionsFile(_latestSdkVersionUrl, "installer");
                 Log.Info($"SDK: {sdkVersion} (Edge)");
             }
             else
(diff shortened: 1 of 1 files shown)


BrennanConroy (approved):

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4153,6 +4153,36 @@ private static async Task<(string version, string hash)> ParseLatestVersionFile(
             }
         }
 
+        /// <summary>
+        /// Parses files that contains lines with [prefix]:[sha], [version]
+        /// </summary>
+        private static async Task<(string version, string hash)> ParseVersionsFile(string urlOrFilename, string prefix)
+        {
+            var content = urlOrFilename.StartsWith("http", StringComparison.OrdinalIgnoreCase)
```
adityamandaleeka, line 4161: Instead of the StartsWith, can we use something like [Uri.IsWellFormedUriString](https://docs.microsoft.com/en-us/dotnet/api/system.uri.iswellformeduristring?redirectedfrom=MSDN&view=net-6.0#System_Uri_IsWellFormedUriString_System_String_System_UriKind_)?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4153,6 +4153,36 @@ private static async Task<(string version, string hash)> ParseLatestVersionFile(
...
+                ? await DownloadContentAsync(urlOrFilename)
+                : await File.ReadAllTextAsync(urlOrFilename)
+                ;
+
+            using (var sr = new StringReader(content))
+            {
+                string line = null;
+
+                while (null != (line = sr.ReadLine()))
```
adityamandaleeka, line 4170: Don't you want this to fail loudly if something is unexpected in the format? e.g. if there are no lines with the prefix or something.

PR #421, opened 2022-03-15
Measure cpustats when limits are sets

sebastienros:
image

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -766,7 +766,7 @@ namespace Microsoft.Crank.Agent
                                             Aggregate = Operation.Max,
                                             Reduce = Operation.Sum,
                                             Format = "n0",
-                                            LongDescription = "Number of period that any thread in the cgroup was runnable",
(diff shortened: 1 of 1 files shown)


adityamandaleeka (approved):

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -5269,6 +5336,44 @@ void SetAttribute(XDocument doc, string path, string name, string value)
...
+
+            using (var sr = new StringReader(result.StandardOutput))
+            {
+                var line = sr.ReadLine();
+
+                while (line != null)
+                {
+                    if (line.StartsWith("nr_periods"))
+                    {
+                        long.TryParse(line.Split(' ', 2, StringSplitOptions.RemoveEmptyEntries).Last(), out nrPeriods);
```
davidfowl, line 5357: defensive parsing?

sebastienros: `TryParse`, `2` and `Last` just in case of unexpected valuesm then it would be `0` which is better than something I don't know can happen.

Did you want to suggest something or just wondering?

PR #434, opened 2022-04-12
Read all profiles before checking service names

sebastienros:


src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -1658,6 +1658,10 @@ int interval
             // 1- The default values of the profile is applied to all jobs
             // 2- The job specific values are applied accordingly to each named job
 
+            // Merge all requested profiles to have the information of all agents available in a single one
+
```
HaoK, line 1662: nit: extra blank line?

HaoK (approved):

PR #435, opened 2022-04-12
Fix duplicate session argument

sebastienros:


src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -159,7 +159,7 @@ public static int Main(string[] args)
             _elastiSearchUrlOption = app.Option("--es",
             "Elasticsearch server url to store results in", CommandOptionType.SingleValue);
             _elasticSearchIndexOption = app.Option("--index",
-                    "Index name of the Elasticsearch server to store results in", CommandOptionType.SingleValue); _sessionOption = app.Option("--session", "A logical identifier to group related jobs.", CommandOptionType.SingleValue);
+                    "Index name of the Elasticsearch server to store results in", CommandOptionType.SingleValue);
```
halter73, line 162:
```suggestion
                    "Index name of the Elasticsearch server to store results in", CommandOptionType.SingleValue);
```

halter73 (approved):

src/Microsoft.Crank.Controller/Program.cs
sebastienros: Thanks, I agree but will do it later, I don't want to start a rebuild since it's slow and I need it to unblock me

PR #445, opened 2022-04-27
Implement based job override on command line

sebastienros:


sebastienros:
Sorry for the unrelated changes.

Goal is to be able to change what application we want to run. Useful when a scenario uses wrk but we want to use bombardier without creating a duplicate scenario. For instance if you need to use a rate, but wrk doesn't support it. Job properties need to be common though (they are).

src/Microsoft.Crank.Controller/Microsoft.Crank.Controller.csproj
```diff
@@ -10,6 +10,7 @@
     <OutputType>Exe</OutputType>
     <Authors>Microsoft</Authors>
     <PackageId>Microsoft.Crank.Controller</PackageId>
+    <LangVersion>9.0</LangVersion>
```
davidfowl, line 13: ?

src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -1696,6 +1695,19 @@ int interval
...
+                    foreach (var argument in arguments.ToArray())
+                    {
+                        if (argument.Key == $"{serviceName}.job")
+                        {
+                            jobName = argument.Value;
+
+                            // Remove this argument
+                            arguments = arguments.Where(x => x.Key != argument.Key);
+                        }
+                    }
```
davidfowl, line 1709: This looks inefficient. 😄

sebastienros: Lame comment.

src/Microsoft.Crank.Controller/Configuration.cs
```diff
@@ -53,7 +53,7 @@ public class Configuration
 
     }
 
-    public class Scenario
+    public class Service
```
davidfowl, line 56: Why this rename?

sebastienros: when I read my code I was confused, in the yml we have scenarios, but this class represents a named service

PR #446, opened 2022-04-27
Store all latencies in milliseconds

sebastienros:


sebastienros:
This messes up with historical results, but right now wrk and bombardier wouldn't return comparable values for the same metric (http/latency/*)

davidfowl (approved):

PR #447, opened 2022-05-12
Don't resolve configuration path if it's an url

kevingosse:
Because of a network error, I had a run fail with the message:

```
Configuration '/mnt/c/git/dd-trace-dotnet/tracer/build/crank/https:/raw.githubusercontent.com/dotnet/crank/main/src/Microsoft.Crank.Jobs.Bombardier/bombardier.yml' could not be loaded.
```
which is incredibly confusing, and had me waste a few minutes to understand why Crank was apparently trying to convert the URL into a path.

This PR fixes the logic to resolve the path in the error message only if it isn't an URL.

### DIFF
--- a/src/Microsoft.Crank.Controller/Program.cs
+++ b/src/Microsoft.Crank.Controller/Program.cs
@@ -2060,10 +2060,12 @@ namespace Microsoft.Crank.Controller
             {
                 string configurationContent;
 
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

sebastienros:
Thanks

PR #448, opened 2022-05-13
Update msquic on linux for .NET 6 and .NET 7

CarnaViire:
MsQuic v2+ is incompatible with MsQuic v1.
.NET 6 needs MsQuic 1.8/1.9, while .NET 7 already depends on MsQuic 2+, so both versions have to be installed on the machine.
Unfortunately, we don't have a signed package for MsQuic 2.0 yet, so the workaround is to build it from sources (we do the same in our CI).

I verified the agent in local docker, that HTTP/3 benchmarks would run on both .NET 6 and latest .NET 7.

cc @ManickaP

### DIFF
--- a/docker/agent/Dockerfile
+++ b/docker/agent/Dockerfile
@@ -40,8 +40,32 @@ RUN apt-get update \
 
 # Install HTTP/3 support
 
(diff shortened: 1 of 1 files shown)


ManickaP (approved):
LGTM, thanks!

sebastienros (approved):

sebastienros:
Should be deployed automatically tonight. I can update aspnet-perf-lin manually if you want to test

PR #449, opened 2022-05-19
Trim regression titles

sebastienros:


sebastienros:
In case the yaml template contains newlines.

adityamandaleeka (approved):

PR #456, opened 2022-08-18
Update NuGet.config for dotnet8 feeds

sebastienros:


NuGet.config
```diff
@@ -7,6 +7,8 @@
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <!-- products -->
+    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
```
wtgodbe, line 10: Do you still use stuff from the dotnet5/dotnet6 feeds? You might be able to get rid of those

wtgodbe (approved):

NuGet.config
sebastienros: will probably get rid of 2.1/3.1 stuff first, 5.0 probably when we start benchmarking 8.0, but keeping 6.0 definitely, still used daily until I switch all baselines to 7.0 in a few months.

PR #458, opened 2022-08-22
Fix Regression.Change to be current - previous

adityamandaleeka:
This changes `Change` to mean the delta between the current result and the one before it.

The issue filer uses this value when filing issues.

@sebastienros


sebastienros:
The log right after is using the same thing that you changed, so I assume it's right. `value3` is the delta from the 1st value to the one with the regression. This should fix the reported issue templates, though one could argue that the actual "change" could be based on the 1st value, since it find regressions if from the 1st value there are two "in-range" ones, then two "out-of-range". So the sample previous to the regression might not impact the "change". I assume we could find some examples where the change would be very small, though really out of range from the 1st value.

sebastienros (approved):

PR #459, opened 2022-08-24
Use 7.0.1xx SDK

sebastienros:


sebastienros:
To prevent performance differences from having us question the SDK version for now.

sebastienros:
Not merging until we know that 8.0 sdk is a problem.

PR #460, opened 2022-08-25
Resolve edge runtime version for internal feeds

sebastienros:


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4017,21 +4023,30 @@ private static async Task<string> ResolveRuntimeVersion(string buildToolsPath, s
 
                 var versionPrefix = targetFramework.Substring(targetFramework.Length - 3);
 
-                foreach (var runtimeApiUrl in _runtimeFeedUrls)
+                if (versionPrefix == "8.0")
```
adityamandaleeka, line 4026: Feels like these blocks are all the same and can be factored out?

adityamandaleeka (approved):
LGTM other than the minor suggestion.

PR #466, opened 2022-09-14
[crank-pr] Use authenticated API calls when creds are available

sebastienros:


sebastienros:
Also paginates PR fetches

src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -117,7 +120,7 @@ static async Task<int> Main(string[] args)
             rootCommand.Handler = CommandHandler.Create<BotOptions>(Controller);
 
             // Parse the incoming args and invoke the handler
-            return await rootCommand.InvokeAsync(args);
+            return await rootCommand.InvokeAsync(args);
```
BrennanConroy, line 123: super nit:
```suggestion
            return await rootCommand.InvokeAsync(args);
```

src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -142,13 +145,21 @@ private static async Task<int> Controller(BotOptions options)
                 return 1;
             }
 
+            CommentCutoffDate =  DateTime.Now.AddMinutes(0 - _options.Age);
```
BrennanConroy, line 148: does `AddMinutes(-_options.Age);` work? Or even `SubtractMinutes(_options.Age);`

src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -142,13 +145,21 @@ private static async Task<int> Controller(BotOptions options)
...
+
             if (_options.GitHubBaseUrl != null)
             {
                 // GitHub Enterprise may require the client to be authenticated,
                 // rather than just use a different base address for the API.
                 await UpgradeAuthenticatedClient();
             }
 
+            // If authentication information is provided pre-authenticate since it will prevent some rate limiting exception for read-only API calls.
+            if (!String.IsNullOrEmpty(_options.AccessToken) || !String.IsNullOrEmpty(_options.AppId) || !String.IsNullOrEmpty(_options.AppId))
```
BrennanConroy, line 158: Should this be an `else if` since the previous `else` might have authenticated already?

nit: Use lowercase `string`

src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -108,7 +108,10 @@ static async Task<int> Main(string[] args)
                     "Any additional arguments to pass through to crank."),
                 new Option<string>(
                     "--config",
-                    "The path to a configuration file.") { IsRequired = true }
+                    "The path to a configuration file.") { IsRequired = true },
+                new Option<int>(
+                    "--age",
```
BrennanConroy, line 113: Are there any docs that need updating with the new option?

src/Microsoft.Crank.PullRequestBot/Program.cs
sebastienros: UpgradeAuthentication can be called multiple times, it's done later on in the code if we need to create new comments for instances, independently of these blocks.

src/Microsoft.Crank.PullRequestBot/Program.cs
sebastienros: Yes :/

src/Microsoft.Crank.PullRequestBot/Program.cs
BrennanConroy: Sure, but why call it right after it was just called? Will that actually change anything?

BrennanConroy (approved):

src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -142,12 +145,19 @@ private static async Task<int> Controller(BotOptions options)
...
 
+            CommentCutoffDate =  DateTime.Now.Subtract(TimeSpan.FromMinutes(_options.Age));
+
             if (_options.GitHubBaseUrl != null)
             {
                 // GitHub Enterprise may require the client to be authenticated,
                 // rather than just use a different base address for the API.
                 await UpgradeAuthenticatedClient();
             }
+            else if (!string.IsNullOrEmpty(_options.AccessToken) || !string.IsNullOrEmpty(_options.AppId) || !string.IsNullOrEmpty(_options.AppId))
```
martincostello, line 156: Did you mean to check something else here? `AppId` is checked for empty twice.

sebastienros: Safer, the value might change between two calls

PR #467, opened 2022-09-14
Add support for multiple pipelines

sebastienros:
With this change only the profiles/benchmarks/components that are available in the configuration file will be used. The goal is to be able to create as many configuration file as there are components/profiles compatible combinations and have different pipelines use each file.

Example:
For the runtime repository benchmarks can target windows/linux and x64/arm64. The idea is to create 4 files, each of them will have the scripts required for a shell/arch couple, and the profiles it supports.

Running a benchmark with "coreclr aspnet-arm-lin plaintext" will then only work for the configuration file that contains the arm/lin commands, using bash as the shell and targeting linux.

If a command contains "coreclr aspnet-arm-lin,aspnet-arm-win plaintext" then one pipeline will be able to run the lin one and another pipeline will run the win one. Even though there is a single command.

### DIFF
--- a/src/Microsoft.Crank.PullRequestBot/Program.cs
+++ b/src/Microsoft.Crank.PullRequestBot/Program.cs
@@ -286,13 +286,6 @@ namespace Microsoft.Crank.PullRequestBot
 
                     try
                     {
-                        if (!HasThingsToRun(command.Benchmarks, command.Profiles, command.Components))
-                        {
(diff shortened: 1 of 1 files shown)


kunalspathak (approved):
I don't understand much of the code but LGTM. Did you try locally and see if it works?

BrennanConroy (approved):

PR #472, opened 2022-10-04
Detect benchmarks downtime

sebastienros:


sebastienros:
Tested locally by trying to detect EF was down, and recovered.

roji (approved):
Looks great, it even has fancy terms like "standard deviation"!

src/Microsoft.Crank.RegressionBot/SourceSection.cs
```diff
@@ -2,13 +2,14 @@
...
 // See the LICENSE file in the project root for more information.
 
-using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Crank.RegressionBot
 {
     public class SourceSection
     {
+        public bool HealthCheck { get; set; } = false;
```
adityamandaleeka, line 11: Comment?

adityamandaleeka (approved):

PR #474, opened 2022-10-06
Fix 7.0 preview packages discovery

sebastienros:
@martincostello @BrennanConroy

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -5561,8 +5561,8 @@ namespace Microsoft.Crank.Agent
             const string publicFeed = "https://dotnetcli.azureedge.net/dotnet";
             
             var dotnetFeeds = version.StartsWith("8.0")
-                ? new string[] { internaFeed, publicFeed } // for 8.0 we check on the internal feed first
-                : new string[] { publicFeed, internaFeed } // for older versions odds are that we are looking for a public package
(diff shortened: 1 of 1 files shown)


sebastienros:
When requesting preview versions of 7.0 explicitly it fails to find the package

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -5562,52 +5553,59 @@ private static async Task<string> GetLatestPackageVersion(string packageIndexUrl
...
-                PackageTypes.AspNetCore => "{3}/aspnetcore/Runtime/{0}/aspnetcore-runtime-{0}-{1}.{2}",
-                PackageTypes.NetCoreApp => "{3}/Runtime/{0}/dotnet-runtime-{0}-{1}.{2}",
-                PackageTypes.WindowsDesktop => "{3}/Runtime/{0}/windowsdesktop-runtime-{0}-{1}.{2}",
-                _ => throw new InvalidOperationException()
-            };
-                
-            var extension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
-                ? "zip"
-                : "tar.gz"
+            const string internaFeed = "https://dotnetbuilds.azureedge.net/public";
```
BrennanConroy, line 5560: nit:
```suggestion
            const string internalFeed = "https://dotnetbuilds.azureedge.net/public";
```

BrennanConroy (approved):

PR #487, opened 2022-10-17
Fix net8.0 aspnet version resolution

sebastienros:


sebastienros:
Missing an `else`, so changed to `switch` to prevent it again

BrennanConroy (approved):

PR #511, opened 2022-12-07
Fix percentiles

eerhardt:
There was a copy-paste error where percentile75 and percentile50 were using `90` for the percentiles.

cc @sebastienros @mitchdenny


sebastienros:
I have looked at this section 50 times, never noticed it :/

sebastienros (approved):

Issue #513, opened 2022-12-09
[Proposal] Accept glob filter to calculate the set of files used for determining Published Size (KB)


mitchdenny:
I think it would be good to be able to filter the set of files that are used to determine the `Published Size (KB)` metric. In my scenario I want to exclude certain files from the calculation which a user of the runtime wouldn't deploy to their production workloads. Some mechanism like this would be handy.

```yaml
scenarios:

  aotstage1:
    application:
      job: aspnetbenchmarks
      selfContained: true
      publishedSizeFilter:
        exclude:
          - !*.dbg
          - !*.json
      buildArguments: 
        - "/p:PublishAot=true"
        - "/p:StripSymbols=true"
        - "/p:DebugSymbols=false"
        - "/p:DebugType=None"
      variables:
        scenario: aotstage1
```

mitchdenny:
`StripSymbols` really relates to _where_ the symbols are placed, not whether they are present or not. In the case of Windows, symbols have always been split into a separate file (`pdb`). In Linux however the symbols by default are baked into the executable. All `StripSymbols` does on Linux is put them into a separate file. (a `dbg` file).

I think most folks that are super concerned about the disk footprint are probably already going to be jettisoning anything like debug symbols via their build system. My ask here is really around being able to narrow down the scope to the files that we expect the developers to care about.

mitchdenny:
... introducing another MSBuild property would work too.

sebastienros:
Crank now records symbols size natively

PR #514, opened 2022-12-12
Record symbols size

sebastienros:
Recording both actual publish size (same as today) and symbols size individually.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -3398,7 +3398,7 @@ namespace Microsoft.Crank.Agent
                 var fis = d.GetFiles();
                 foreach (var fi in fis)
                 {
-                    if (ignoredExtensions != null && ignoredExtensions.Contains(fi.Extension))
+                    if (ignoredExtensions != null && ignoredExtensions.Contains(fi.Extension, StringComparer.OrdinalIgnoreCase))
                     {
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


sebastienros:
Fixes #513

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -752,6 +753,19 @@ private static async Task ProcessJobs(string hostname, string dockerHostname, Ca
                                         });
                                     }
 
+                                    if (!job.Metadata.Any(x => x.Name == "benchmarks/symbols-size"))
+                                    {
+                                        job.Metadata.Enqueue(new MeasurementMetadata
+                                        {
+                                            Source = "Host Process",
+                                            Name = "benchmarks/symbols-size",
```
mkArtakMSFT, line 761: Looks like you use the same text here as on line 756. I'd just move it into a constant and reuse in both places.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3206,6 +3220,18 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
                 });
             }
 
+            var publishedSizeWithoutSymbols = DirSize(new DirectoryInfo(outputFolder), _ignoredSymbolsExtensions) / 1024;
+
+            if (publishedSize != 0 && publishedSizeWithoutSymbols != 0)
+            {
+                job.Measurements.Enqueue(new Measurement
+                {
+                    Name = "benchmarks/symbols-size",
```
mkArtakMSFT, line 3229: 3rd occurrence 👍

mkArtakMSFT (approved):

PR #515, opened 2022-12-12
Prevent download/list action from accessing files outside of the job

sebastienros:


src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -645,7 +651,13 @@ public IActionResult List(int id, string path)
                     return NotFound();
                 }
 
-                var fullPath = Path.Combine(job.BasePath, path);
+                // Resolve dot notation in path
+                var fullPath = Path.GetFullPath(path, job.BasePath);
```
mkArtakMSFT, line 655: Would a tiny helper method be useful here? Looks like you have the very same check in both places.

mkArtakMSFT (approved):
Just one small suggestion.

PR #516, opened 2022-12-13
Send output files as a single zip

sebastienros:
Instead of single individual files using gzip.

### DIFF
--- a/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
+++ b/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
@@ -357,20 +357,25 @@ namespace Microsoft.Crank.Agent.Controllers
                 return StatusCode(500, $"The job can't accept attachment as its state is {job.State}");
             }
 
+            var destinationTempFilename = Path.GetFullPath(Path.GetRandomFileName(), Path.GetTempPath());
+
             var tempFilename = Path.GetTempFileName() + ".zip";
 
             await SaveBodyAsync(tempFilename);
 
             job.LastDriverCommunicationUtc = DateTime.UtcNow;
 
-            var destinationTempFilename = Path.GetFullPath(Path.GetRandomFileName(), Path.GetTempPath());
-            
-            // Extract the zip file in a temporary folder
-            ZipFile.ExtractToDirectory(tempFilename, destinationTempFilename);
+            try
+            {
+                // Extract the zip file in a temporary folder
+                ZipFile.ExtractToDirectory(tempFilename, destinationTempFilename);
+            }
+            finally
+            {
+                System.IO.File.Delete(tempFilename);
+            }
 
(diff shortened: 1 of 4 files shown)


trylek:
Thanks @sebastienros for making the change; it looks great to me, I have approved it but apparently I don't have write access to the crank repo so that the automated merge didn't happen; I'm afraid you might need to identify another reviewer with write access.

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -332,6 +332,64 @@ public async Task<IActionResult> UploadAttachment(int id)
...
+
+            if (job == null)
+            {
+                return NotFound("Job doesn't exist anymore");
+            }
+
+            if (job.State != JobState.Initializing)
+            {
+                Log.Info($"Attachment rejected, job is {job.State}");
+                return StatusCode(500, $"The job can't accept attachment as its state is {job.State}");
```
mkArtakMSFT, line 357: I agree that it's worth logging the current state of the job, but returning it to the caller may not be a good idea. Maybe simply respond that it's not in the `initializing` state and do not disclose the current state?

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -332,6 +332,64 @@ public async Task<IActionResult> UploadAttachment(int id)
...
+            job.LastDriverCommunicationUtc = DateTime.UtcNow;
+
+            var destinationTempFilename = Path.GetFullPath(Path.GetRandomFileName(), Path.GetTempPath());
+            
+            // Extract the zip file in a temporary folder
+            ZipFile.ExtractToDirectory(tempFilename, destinationTempFilename);
+
+            job.LastDriverCommunicationUtc = DateTime.UtcNow;
+
+            System.IO.File.Delete(tempFilename);
```
mkArtakMSFT, line 373: What if some unexpected error happens above? I'd suggest wrapping lines 362-371 in a try - finally block and moving the `Delete` into a finally block.

mkArtakMSFT (changes requested):
Just one small change to avoid a situation where the storage gets filled up.

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
sebastienros: It could be important for debugging to understand at what point the client tried to upload a file and how we got into this state. There is nothing to hide here.

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
sebastienros: Done

mkArtakMSFT (approved):

PR #517, opened 2022-12-15
Introduce custom network constants 

sebastienros:
Improves reliability on slow networks.


src/Microsoft.Crank.Controller/JobConnection.cs
```diff
@@ -34,6 +34,9 @@ public class JobConnection
 
         private string _serverJobUri;
         private bool _keepAlive;
+        private int _keepAliveTimeoutMilliseconds = 5000;
+        private int _keepAlivePeriodMilliseconds = 2000;
+        private int _uploadBufferSize = 1024 * 1204; // default is 80K which is too small on slow network connections
```
adityamandaleeka, line 39: 1204?

sebastienros: It's better than 1024

adityamandaleeka (approved):

PR #519, opened 2022-12-17
Delete attachments when job fails

sebastienros:


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -1696,7 +1696,16 @@ async Task DeleteJobAsync()
 
                                     // Delete application folder
 
-                                    await TryDeleteDirAsync(tempDir, false);
+                                    await TryDeleteDirAsync(tempDir);
+                                }
+
+                                // Delete temporary attachment files
```
davidfowl, line 1702: Should this be in a finally?

sebastienros: Done

davidfowl (approved):

PR #522, opened 2022-12-29
Improve .gitignore support

sebastienros:
Found a bug while working on a local folder benchmark: `/obj` would be uploaded disregarding the `[Oo]bj/` rule, so implemented some tests based on the spec https://git-scm.com/docs/gitignore and fixed other issues.

### DIFF
--- a/src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
+++ b/src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Crank.Controller.Ignore
             
             while (currentDir != null && currentDir.Exists)
             {
-                var gitIgnoreFilename = Path.GetFullPath(currentDir.FullName + "/.gitignore");
+                var gitIgnoreFilename = Path.GetFullPath(Path.Combine(currentDir.FullName, ".gitignore"));
 
                 if (File.Exists(gitIgnoreFilename))
                 {
@@ -35,7 +35,7 @@ namespace Microsoft.Crank.Controller.Ignore
                     var localRules = new List<IgnoreRule>();
 
                     // Don't process parent folder if we are at the repository level
-                    if (Directory.Exists(Path.GetFullPath(currentDir.FullName + "/.git")))
+                    if (Directory.Exists(Path.GetFullPath(Path.Combine(currentDir.FullName, ".git"))))
                     {
                         currentDir = null;
                     }
--- a/src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
+++ b/src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
@@ -65,13 +65,18 @@ namespace Microsoft.Crank.Controller.Ignore
                             }
 
                             // Trailing spaces are ignored unless they are quoted with backslash ("\").
-                            if (rule.Contains(@"\ "))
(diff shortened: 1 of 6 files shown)


src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
```diff
@@ -26,25 +20,22 @@ private IgnoreFile(string gitIgnorePath)
...
-            var ignoreFile = new IgnoreFile(path);
+            var ignoreFile = new IgnoreFile();
 
-            var currentDir = new DirectoryInfo(Path.GetDirectoryName(path));
+            var currentDir = Directory.Exists(path) ? new DirectoryInfo(path) : new DirectoryInfo(Path.GetDirectoryName(path));
             
             while (currentDir != null && currentDir.Exists)
             {
-                var gitIgnoreFilename = Path.Combine(currentDir.FullName, ".gitignore");
+                var gitIgnoreFilename = Path.GetFullPath(currentDir.FullName + "/.gitignore");
```
BrennanConroy, line 29: Shouldn't this still be a Path.Combine?

src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
```diff
@@ -26,25 +20,22 @@ private IgnoreFile(string gitIgnorePath)
...
                     var basePath = currentDir.FullName.Replace("\\", "/") + "/";
 
                     var localRules = new List<IgnoreRule>();
 
-                    // Ignore .git folders by default
-                    localRules.Add(IgnoreRule.Parse(basePath, ".git/"));
-
                     // Don't process parent folder if we are at the repository level
-                    if (Directory.Exists(Path.Combine(currentDir.FullName, ".git")))
+                    if (Directory.Exists(Path.GetFullPath(currentDir.FullName + "/.git")))
```
BrennanConroy, line 38: Same

src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
```diff
@@ -67,6 +58,28 @@ public static IgnoreFile Parse(string path, bool includeParentDirectories = fals
                                 continue;
                             }
 
+                            // Put a backslash ("\") in front of the first hash for patterns that begin with a hash.
```
BrennanConroy, line 61: Nit: this comment looks like it was copied from the spec, our code removes the `\` which seems odd without context.
TLDR; add/update comment
Same in IgnoreRule.cs

src/Microsoft.Crank.Controller/Ignore/IgnoreFile.cs
sebastienros: I preferred `GetGullPath` because 
1- that's what is required, it create the fully qualified path
2- if the second argument starts with `/` it won't root it to the drive

So maybe here it's the same, but I tried to use the same pattern instead of mixing solutions

BrennanConroy: I meant
```
Path.GetFullPath(Path.Combine(currentDir.FullName, ".gitignore"));
```

sebastienros: Should work here. I had a unit test failing because of that but maybe it was for another place.

BrennanConroy (approved):

PR #525, opened 2023-01-10
Add option to ignore .gitignore files

sebastienros:


src/Microsoft.Crank.Controller/JobConnection.cs
```diff
@@ -943,7 +943,10 @@ private static void DoCreateFromDirectory(string sourceDirectoryName, string des
             {
                 var basePath = di.FullName;
 
-                var ignoreFile = IgnoreFile.Parse(Path.Combine(sourceDirectoryName, ".gitignore"));
+                var ignoreFile = Job.Options.NoGitIgnore
+                    ? IgnoreFile.Parse(Path.Combine(sourceDirectoryName, ".gitignore"))
+                    : new IgnoreFile()
+                    ;
```
brunolins16, line 949: Should not be:


```suggestion
                var ignoreFile = Job.Options.NoGitIgnore
                    ? new IgnoreFile()
                    : IgnoreFile.Parse(Path.Combine(sourceDirectoryName, ".gitignore"))
                    ;
```

brunolins16: Or (to follow what the parameter description says) :

```suggestion
                var ignoreFile = !Job.Options.NoGitIgnore
                    ? IgnoreFile.Parse(Path.Combine(sourceDirectoryName, ".gitignore"))
                    : new IgnoreFile()
                    ;
```

brunolins16 (approved):

PR #526, opened 2023-01-12
Add support for metrics

sebastienros:


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3061,71 +3061,20 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
...
-                //$"/p:BenchmarksNETStandardImplicitPackageVersion={aspNetCoreVersion} " +
-                //$"/p:BenchmarksNETCoreAppImplicitPackageVersion={aspNetCoreVersion} " +
-                //$"/p:BenchmarksRuntimeFrameworkVersion={runtimeVersion} " +
-                //$"/p:BenchmarksTargetFramework={targetFramework} " +
-                //$"/p:BenchmarksAspNetCoreVersion={aspNetCoreVersion} " +
-                //$"/p:MicrosoftAspNetCoreAllPackageVersion={aspNetCoreVersion} " +
-                //$"/p:NETCoreAppMaximumVersion=99.9 "; // Force the SDK to accept the TFM even if it's an unknown one. For instance using SDK 2.1 to build a netcoreapp2.2 TFM.
+                $"/p:MicrosoftAspNetCoreAppPackageVersion={aspNetCoreVersion} " +
+                $"/p:GenerateErrorForMissingTargetingPacks=false " +
+                $"/p:RestoreNoCache=true " // #1445 force no cache for restore to avoid restore failures for packages published within last 30 minutes
```
joperezr, line 3066: > #1445

Is that referencing an issue in some repo that is tracking the restore issue? I looked in dotnet/aspnetcore but the issue didn't seem to be related.

joperezr: Oh, I now see that you just copied that from bellow. If this is a stale comment, I guess we could technically delete it right?

sebastienros: Fixed the link

joperezr (approved):
Not very familiar with the source code, but changes LGTM

PR #532, opened 2023-02-01
Compress regression data and limit issues per report

sebastienros:


src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -538,7 +542,7 @@ private static async IAsyncEnumerable<Regression> FindRegression(Source source)
                         var result = new BenchmarksResult
                         {
                             Id = Convert.ToInt32(reader["Id"]),
-                            Excluded = Convert.ToBoolean(reader["Excluded"]),
+                            Excluded = Convert.ToBoolean(reader["Excluded"] as bool? ?? false), // Handle DBNull values
```
BrennanConroy, line 545: nit: the `Convert` call doesn't look like it's needed

src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -359,19 +366,16 @@ private static async Task CreateRegressionIssue(IEnumerable<Regression> regressi
 
             var title = await CreateIssueTitle(regressions, titleTemplate);
 
-            if (!_options.Debug)
```
BrennanConroy, line 362: Is this not helpful to have?

src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -297,6 +299,11 @@ private static async Task<string> CreateIssueBody(IEnumerable<Regression> regres
 
                 body += regressionBlock;
 
+                if (body.Length > 65536)
+                {
+                    throw new Exception($"Body too long ({body.Length} > 65536 chars)");
```
BrennanConroy, line 304: Any way to try truncating first? I assume this error shows up in the regression bot AzDo run? Does it send an email if it fails?

src/Microsoft.Crank.RegressionBot/Program.cs
sebastienros: Fixed

src/Microsoft.Crank.RegressionBot/Program.cs
sebastienros: No, it's hiding too much, could repro a bug with it, and without it would do too few.

src/Microsoft.Crank.RegressionBot/Program.cs
sebastienros: Can truncate because it's a binary payload for serialized data, so couldn't be deserialized and the issue would not be tracked (the issues would be recreated every time they are detected).

I need to create another PR that will actually send a != 0 exit code such that Azdo knows the job is failing. It's currently green. Need to check how to handle that with System.CommandLine

BrennanConroy (approved):

src/Microsoft.Crank.RegressionBot/Program.cs
sebastienros: Ideally issues would be filled with regressions until there is no more space. And have a block of MSGPACK per issue then create a new page, the idea being to limit the number of GH issues created.

I was surprised of the final size of the serialized text. Maybe Base64 is not optimal here. But even from MsgPack the size is significant, and I checked only the required data is serialized.

PR #538, opened 2023-02-02
Update msquic setup support in docker file

sebastienros:
Build is currently broken, using package instead since they are now available.


CarnaViire (approved):
Thanks!

wtgodbe (approved):

PR #544, opened 2023-02-16
Fix payload parsing

sebastienros:


src/Microsoft.Crank.AzureDevOpsWorker/JobPayload.cs
```diff
@@ -40,7 +38,7 @@ public static JobPayload Deserialize(byte[] data)
                 var index = -1;
                 do
                 {
-                    index = str.IndexOf('{', index);
+                    index = str.IndexOf('{', index + 1);
```
BrennanConroy, line 41: What's the while loop for? Don't you just want to find the first `{`?

sebastienros: The first one that is followed by a valid char (WS or ") because if you look at the comments i nthe test the payload is formatted using DataContracSerializer and contains extra one, not always followed by the same thing.

sebastienros: https://github.com/dotnet/crank/pull/544/files#diff-61d67cb24c2a9c0a529927e2784b4f068ac030333aec3a5b22cf101e48cec028R21

BrennanConroy: Oh, looked closer at the example and it has an added `{`...

BrennanConroy (approved):

PR #545, opened 2023-02-17
Allow project relative file downloads

sebastienros:
Necessary to download files from the csproj folder or `obj/,...`

### DIFF
--- a/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
+++ b/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
@@ -658,7 +658,7 @@ namespace Microsoft.Crank.Agent.Controllers
                 var rootPath = Directory.GetParent(job.BasePath).FullName;
 
                 // Assume ~/ means relative to the project, otherwise it's relative to the application folder (job.BasePath)
-                var isRelativeToProject = String.IsNullOrEmpty(job.Source.DockerFile) && path.StartsWith("~/") || path.StartsWith("~\\");
+                var isRelativeToProject = String.IsNullOrEmpty(job.Source.DockerFile) && (path.StartsWith("~/") || path.StartsWith("~\\"));
                 
                 if (isRelativeToProject)
                 {
@@ -732,7 +732,7 @@ namespace Microsoft.Crank.Agent.Controllers
                 var rootPath = Directory.GetParent(job.BasePath).FullName;
 
                 // Assume ~/ means relative to the project, otherwise it's relative to the application folder (job.BasePath)
-                var isRelativeToProject = String.IsNullOrEmpty(job.Source.DockerFile) && path.StartsWith("~/") || path.StartsWith("~\\");
+                var isRelativeToProject = String.IsNullOrEmpty(job.Source.DockerFile) && (path.StartsWith("~/") || path.StartsWith("~\\"));
 
                 if (isRelativeToProject)
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Crank.Controller/JobConnection.cs
```diff
@@ -1218,6 +1218,11 @@ public async Task DownloadFileAsync(string file, string output)
                 var uri = Combine(_serverJobUri, "/download?path=" + HttpUtility.UrlEncode(file));
                 Log.Verbose("GET " + uri);
 
+                if (file.StartsWith("~/") || file.StartsWith("~\\"))
```
eerhardt, line 1221: Does `~/` work when the path has a `*`?

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -654,16 +654,29 @@ public async Task<IActionResult> Download(int id, string path)
                     return NotFound();
                 }
 
+                // Downloads can't get out of this path
+                var rootPath = Directory.GetParent(job.BasePath).FullName;
+
+                // Assume ~/ means relative to the project, otherwise it's relative to the application folder (job.BasePath)
+                var isRelativeToProject = String.IsNullOrEmpty(job.Source.DockerFile) && path.StartsWith("~/") || path.StartsWith("~\\");
```
eerhardt, line 661: Do you need parens here?

```
String.IsNullOrEmpty(job.Source.DockerFile) && (path.StartsWith("~/") || path.StartsWith("~\\"));
```

eerhardt, line 661: Why the check for if the DockerFile is empty? Does that mean it doesn't work with docker files?

src/Microsoft.Crank.Controller/JobConnection.cs
sebastienros: Filenames here don't have `*`, the controller invokes a /list first when there is stars. Then invokes individual calls to /download

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
sebastienros: yes

sebastienros: Not sure what it would mean for docker file, so I want to ignore it for now. Foo csprojs I know that a sub folder named 'published' is created, and it's the default location where we run apps. But for Docker it depends on each image what the current path is.

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -715,11 +728,24 @@ public IActionResult List(int id, string path)
...
```
eerhardt, line 735: same comments as above

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -654,16 +654,29 @@ public async Task<IActionResult> Download(int id, string path)
...
+                
+                if (isRelativeToProject)
+                {
+                    // One level up from the output folder
+                    path = string.Concat("..", path.AsSpan(1));
```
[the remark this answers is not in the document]
sebastienros: no, `~/obj/file.txt` is translated to `../obj/file.txt`

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
sebastienros: yep

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
eerhardt: Should we throw in that case then? To tell the user that this isn't possible?

eerhardt (approved):
Just one more comment about throwing with a docker file + `~` option.

PR #546, opened 2023-02-17
Add file manipulation methods in scripts

sebastienros:


src/Microsoft.Crank.Controller/ScriptFile.cs
```diff
@@ -0,0 +1,37 @@
...
+            }
+
+            return File.ReadAllText(filename);
+        }
+
+        public void WriteFile(string filename, string data)
+        {
+            if (String.IsNullOrEmpty(filename))
+            {
+                return;
```
eerhardt, line 26: throw? that way people know what they are trying to do isn't working.

src/Microsoft.Crank.Controller/ScriptFile.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.IO;
+
+namespace Microsoft.Crank.Controller
+{
+    public class ScriptFile
+    {
+        public string ReadFile(string filename)
+        {
+            if (String.IsNullOrEmpty(filename))
```
eerhardt, line 14:
```suggestion
            if (string.IsNullOrEmpty(filename))
```

https://blog.paranoidcoding.com/2019/04/08/string-vs-String-is-not-about-style.html

😉

src/Microsoft.Crank.Controller/ScriptConsole.cs
```diff
@@ -13,27 +13,47 @@ public class ScriptConsole
...
         public void Info(params object[] args)
         {
+            if (args == null || args.Length == 0)
+            {
+                return;
+            }
+
             Console.ForegroundColor = ConsoleColor.Green;
-            Console.WriteLine(String.Join(" ", args.Select(x => x.ToString())));
+            Console.WriteLine(String.Join(" ", args.Select(x => x?.ToString())));
```
eerhardt, line 32: Can't you just use `string.Join(string, params object[])` directly? It handles `null`.

eerhardt (approved):

PR #549, opened 2023-02-23
Fix relative path in list action

sebastienros:


src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -754,11 +754,12 @@ public IActionResult List(int id, string path)
                     return Json(Array.Empty<string>());
                 }
 
+                // Returned paths are relative to the job.BasePath folder (e.g., /published in case of dotnet projects)
                 if (fullPath.Contains("*"))
                 {
                     return Json(
                         Directory.GetFiles(Path.GetDirectoryName(fullPath), Path.GetFileName(fullPath))
-                        .Select(x => x.Substring(rootPath.Length).TrimStart('/', '\\'))
+                        .Select(x => x.Substring(job.BasePath.Length).TrimStart('/', '\\'))
```
BrennanConroy, line 762: Shouldn't this be `fullPath`?

sebastienros: `BasePath` is something like `c:\temp\clone\src\published\`. Users send `path` as `folder\*.*` which is relative to `BasePath`. The results need to be something that is also relative to `BasePath` such than the client can then call `/download?path=xyz` with the relative path.

If we were use `fullPath` here it would contain `folder\` and then the results would not be relative to `BasePath` anymore. And `fullPath` also contains a globing pattern.

BrennanConroy (approved):

PR #550, opened 2023-02-24
Don't stop on first healthcheck detection

sebastienros:


src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -922,7 +924,7 @@ private static async IAsyncEnumerable<Regression> FindRegression(Source source)
                             var regression = new Regression
                             {
                                 PreviousResult = results[i - 1],
-                                CurrentResult = results[i],
+                                CurrentResult = results[i - 1],
```
ReubenBond, line 927: It's not clear to me why the current and previous results are identical here.

sebastienros: This model is made for detecting regressions, with Previous/Current, but here I am using the same bot/model to report the fact that the benchmark stopped running. So we can only provide a single result, which is the last one. Another option might be to set PreviousResult to null, it's more logical, but at the same time this could lead to null refs if a template is not expecting that.

BrennanConroy: Add a comment explaining why they have the same result

BrennanConroy (approved):

src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -912,17 +912,23 @@ private static async IAsyncEnumerable<Regression> FindRegression(Source source)
                             Console.WriteLine($"Value: {currentValue}, benchmark runs on average every {(int)TimeSpan.FromSeconds(averageInSeconds).TotalMinutes} minutes with a stdev of {(int)TimeSpan.FromSeconds(standardDeviation).TotalMinutes} minutes. Intervals were: {String.Join(',', intervalsInSeconds)}");
                         }
 
+                        // We assume the benchmark is not running if it wasn't triggered for twice the expected delay.
+                        // The standard deviation could also be ignore here but since it's available let's take it into account.
```
joperezr, line 916:
```suggestion
                        // The standard deviation could also be ignored here but since it's available let's take it into account.
```
NIT

src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -912,17 +912,23 @@ private static async IAsyncEnumerable<Regression> FindRegression(Source source)
...
                         var changeInSeconds = (currentValue.Ticks - previousValue.Ticks) / TimeSpan.TicksPerSecond;
-                        var acceptedChange = (averageInSeconds + 2 * standardDeviation);
+                        var acceptedChange = 2 * (averageInSeconds + standardDeviation);
 
                         var hasRegressed = changeInSeconds > acceptedChange;
 
                         if (hasRegressed)
                         {
+                            // Assign previous and current results to the same value to prevent the current result
+                            // from being empty. Halthchecks don't compare two results, but detect the second
```
joperezr, line 925:
```suggestion
                            // from being empty. Healthchecks don't compare two results, but detect the second
```

joperezr (approved):

Issue #345, opened 2021-10-28
Remove Manatee.Json dependency


sebastienros:
Use https://www.nuget.org/packages/JsonSchema.Net/ instead which is a replacement.

PR #553, opened 2023-02-28
Remove Manatee.Json dependency

catcherwong:
For #345 , use **JsonSchema.Net** to replace **Manatee.Json**.

### DIFF
--- a/src/Microsoft.Crank.PullRequestBot/Microsoft.Crank.PullRequestBot.csproj
+++ b/src/Microsoft.Crank.PullRequestBot/Microsoft.Crank.PullRequestBot.csproj
@@ -28,17 +28,12 @@
     <PackageReference Include="Octokit" Version="0.48.0" />
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
     <PackageReference Include="YamlDotNet" Version="8.1.2" />
-    <PackageReference Include="Manatee.Json" Version="13.0.5" />
     <PackageReference Include="Fluid.Core" Version="2.2.6" />
 
     <!-- Force version for Component Governance compliance -->
-    <PackageReference Include="System.Net.Http" Version="4.3.4">
+    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0">
       <AllowExplicitReference>true</AllowExplicitReference>
     </PackageReference>
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2">
-      <AllowExplicitReference>true</AllowExplicitReference>
-    </PackageReference>
-    
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


catcherwong:
@dotnet-policy-service agree

catcherwong:
Hi, @sebastienros 

It seems that CI can't find the newest version of `JsonSchema.Net`, what can I do for it.

sebastienros:
@catcherwong we are using internal nuget feeds for component governance (security measures), I need to ask our team to add the package to this feed. Will restart the build when it's done.

src/Microsoft.Crank.PullRequestBot/Microsoft.Crank.PullRequestBot.csproj
```diff
@@ -28,17 +28,12 @@
     <PackageReference Include="Octokit" Version="0.48.0" />
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
     <PackageReference Include="YamlDotNet" Version="8.1.2" />
-    <PackageReference Include="Manatee.Json" Version="13.0.5" />
     <PackageReference Include="Fluid.Core" Version="2.2.6" />
 
     <!-- Force version for Component Governance compliance -->
-    <PackageReference Include="System.Net.Http" Version="4.3.4">
+    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0">
```
sebastienros, line 34: This might not even be necessary anymore. I assume STEW was referenced directly by another lib (Manatee?) and then I had to force it to a more recent version. If you could try to build the solution locally without the reference, and it restores something >= 4.3.4 then we can remove this package ref.

sebastienros (commented):
Can you also update the CREDITS file to replace Manatee?

src/Microsoft.Crank.PullRequestBot/Microsoft.Crank.PullRequestBot.csproj
catcherwong: After removing this package ref, it can be successfully built, and all tests have passed. 

image

sebastienros: And do you see what version it restored? I will also follow the version trail to see which one it would use.

catcherwong: **System.Net.Http** was removed.

**System.Text.Encodings.Web** is 6.0.0

sebastienros: Thanks, should be good then. Still waiting for the packages to be published.

catcherwong:
> Can you also update the CREDITS file to replace Manatee?

Should we update here, use JsonSchema.Net to replace Manatee.Json.

sebastienros:
Yes, remove any mention of manatee and its dependencies and add the ones from JsonSchema.Net. Thanks

catcherwong:
> remove any mention of manatee and its dependencies and add the ones from JsonSchema.Net

Done.

LICENSE using JsonSchema.Net's main repository.

https://github.com/gregsdennis/json-everything/blob/master/LICENSE

sebastienros (approved):

PR #561, opened 2023-03-08
Fix duplicate folder name when use downloadFiles options

### LINKED ISSUE #282: DownloadFiles options duplicates folder name
For instance `App_Data/logs/*.*` will download files in `App_Data/logs/App_Data/logs/`

catcherwong:
fixed #282 

------
Using the following command to test

```
crank --config .\samples\local\local.benchmarks.yml --scenario hello --profile local --application.options.downloadFiles App_Data/logs/*.* 
```

Before modification

After modification


sebastienros:
Would you mind trying to add a unit test? I added one recently related to downloading files, so you should be able to copy it and update the conditions.

sebastienros (approved):

PR #563, opened 2023-03-10
Fix regression bot json schema validation

sebastienros:
/cc @catcherwong 

I haven't found why yet ... but that seems to fix the validation


catcherwong:
@sebastienros I'm sorry for this problem.

**JsonSchema.Net** only supports draft 6 and later, update to draft 7 can solve it,

https://json-everything.net/json-schema#schema-versions

PR #569, opened 2023-04-11
Fix cpuSet parsing for a single value

sebastienros:


sebastienros:
@RobertHenry6bev no but it could. The documentation doesn't state that though, and that could be ambiguous in command line arguments if you don't quote the value. So maybe no start trying to do it.

BrennanConroy (approved):

Issue #530, opened 2023-01-31
Add option to breakdown AOT app size


eerhardt:
When publishing an app for NativeAOT, it is very beneficial to get a breakdown of where the app size is coming from.

Here's an example of npgsql doing this in its CI:

https://github.com/npgsql/npgsql/actions/workflows/native-aot.yml (open a successful run and expand the "binary size" details sections.

![image](https://user-images.githubusercontent.com/8291187/215842802-bccbd95f-9b3c-4911-b007-42586466fe14.png)

For NativeAOT apps, the way this information is retrieved is to:

1. Set `<IlcGenerateMstatFile>true</IlcGenerateMstatFile>` MSBuild property
2. Publish the app for NativeAOT
3. In obj\<Configuration>\<Target_Framework>\<RID>\native\ there will be an <AppName>.mstat file. This file contains information about what methods, types, etc. were left in the app. And gives a rough size of the methods in the app.
4. Use some custom C# code to read and gather the information in the format we want to show the dev. Here are some tools that can read this info:
    1. https://gist.github.com/MichalStrehovsky/2c7cb3d623c7f8901541914dab04238d
    2. https://github.com/npgsql/npgsql/blob/218a160e396636f2c56ca7215d264ce644b1464f/test/MStatDumper/Program.cs#L8

We should add an option to crank that will do these steps and send the breakdown of the app size to the developer.

cc @mitchdenny @sebastienros @cincuranet

sebastienros:
Fixed

PR #577, opened 2023-04-22
Add PublishAot option to breakdown AOT app size

catcherwong:
## Add PublishAot option to breakdown AOT app size

- Add option **PublishAot** in `crank`
- Add two measurements about native AOT size in `crank-agent`

Simple sample and result

```
crank --config .\hello.benchmarks.yml --scenario hello --profile local --application.publishAot true --application.framework net7.0 --json res.json
```

ref #530


catcherwong:
```
[14:22:48 ERR] [ERROR] Event pipe session creation timed out. Application might be stopped (application:1)
...
[14:23:05 WRN] [WARNING] No event source open for job 'application' (1)
```
But I don't know why the native aot application cannot `StartEventPipeSession` successfully.

https://github.com/dotnet/crank/blob/1489e624813300f7298de7aed0f03fbdb0198593/src/Microsoft.Crank.Agent/Startup.cs#L4902

sebastienros:
NAtiveAot doesn't support event pipes (for now).

sebastienros:
@eerhardt Can you check if that would be useful and if it needs to do more?

eerhardt:
> Can you check if that would be useful and if it needs to do more?

I don't know what "that" and "it" is in the above sentence.

eerhardt:
Are we able to add the app size measurements without making an explicit `PublishAot` command?

src/Microsoft.Crank.Agent/MstatDumper.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            var mstats = Directory.EnumerateFiles(path, "*.mstat");
+
+            if (mstats == null || !mstats.Any())
+            {
+                return new DumperResults();
+            }
+
+            using var asm = AssemblyDefinition.ReadAssembly(mstats.First());
+            var globalType = (TypeDefinition)asm.MainModule.LookupToken(0x02000001);
```
eerhardt, line 25: @MichalStrehovsky @agocke - did we have a library that was able to parse and read mstat information? Can we use that here?

sebastienros:
@catcherwong I was chatting with Eric offline and we are wondering if we could just omit the publishaot argument altogether and let the crank configuration for the build define these directly so we don't have to add an option. But we could still detect that the mstat file is generated (in the agent) and then add more measurements like you did. Something useful would also be to expose the rest of the information from the mstat file into the json blob result as raw json measurement (same as what the Bombardier load generator is doing).

src/Microsoft.Crank.Agent/MstatDumper.cs
MichalStrehovsky: We don't have one. https://github.com/dotnet/runtimelab/tree/feature/aot-tools is something that could turn into it at some point.

catcherwong:
> if we could just omit the publishaot argument altogether and let the crank configuration for the build define these directly so we don't have to add an option. But we could still detect that the mstat file is generated (in the agent)

OK, I will try to directly check `obj<Configuration><Target_Framework><RID>\native\` directory without adding the new publishaot argument, and if there is an mstat file, calculate the relevant measurements based on this file. If not, ignore these measurements.

>  Something useful would also be to expose the rest of the information from the mstat file into the json blob result as raw json measurement (same as what the Bombardier load generator is doing).

OK.

catcherwong:
When specifying **--application.buildArguments "/p:IlcGenerateMstatFile=true"**, the agent can read the mstat file directly without the publishaot argument. If not specified, there will be no effect.

```sh
crank --config .\local.benchmarks.yml \
    --scenario hello \
    --profile local \
    --application.framework net8.0 \
    --json local-80-win.json \
    --application.packageReferences Microsoft.Dotnet.ILCompiler=8.0.0-preview.3.23174.8 \
    --application.buildArguments "/p:StripSymbols=true" \
    --application.buildArguments "/p:PublishAot=true"  \
    --application.buildArguments "/p:IlcGenerateMstatFile=true" \
    --application.aspNetCoreVersion 8.0.0-preview.3.23177.8 \
    --application.runtimeVersion 8.0.0-preview.3.23174.8 \
    --application.sdkVersion 8.0.100-preview.3.23178.7
```

```sh
crank --config .\local.benchmarks.yml \
    --scenario hello \
    --profile local \
    --application.framework net8.0 \
    --json local-80-win-nomstat.json \
    --application.packageReferences Microsoft.Dotnet.ILCompiler=8.0.0-preview.3.23174.8 \
    --application.buildArguments "/p:StripSymbols=true" \
    --application.buildArguments "/p:PublishAot=true"  \
    --application.aspNetCoreVersion 8.0.0-preview.3.23177.8 \
    --application.runtimeVersion 8.0.0-preview.3.23174.8 \
    --application.sdkVersion 8.0.100-preview.3.23178.7
```

image

It works well in .net 7.0 and .net 8.0 preview 3.  But in .net 8.0 preview 5, it can not read the mstat file correctly.

```
[20:36:21 ERR] Read mstat file [C:\Users\catcherwong\AppData\Local\Temp\2\benchmarks-agent\benchmarks-server-6524\oarour4r.gc5\src\obj\Release\net8.0\win-x64\native\hello.mstat] failed
System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'Mono.Cecil.TypeReference'.
   at Microsoft.Crank.Agent.MstatDumper.GetTypes(MethodDefinition types)+MoveNext() in C:\code\d\crank\src\Microsoft.Crank.Agent\MstatDumper.cs:line 129
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Microsoft.Crank.Agent.MstatDumper.GetInfo(String path) in C:\code\d\crank\src\Microsoft.Crank.Agent\MstatDumper.cs:line 35
```

catcherwong:
@sebastienros  Is there any suggestions for this PR?

sebastienros (approved):

PR #579, opened 2023-04-26
Adding more PR benchmarks log

sebastienros:


src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -868,8 +881,6 @@ private static async Task<IEnumerable<Result>> RunBenchmark(Command command, str
                     Console.WriteLine($"Failed to clean {cloneFolder}. Ignoring...");
                 }
             }
-
-            return results;
```
kunalspathak, line 872: we don't want to return `null` or something here?

sebastienros: No, it never goes there, or the compiler would complain that the method has no return value.

kunalspathak (approved):

BrennanConroy (approved):

PR #580, opened 2023-04-27
Fix thumbprint initialization

sebastienros:
It was not initialized when running the bot so all thumbprints were identical.


src/Microsoft.Crank.PullRequestBot/Program.cs
```diff
@@ -170,6 +170,8 @@ private static async Task<int> Controller(BotOptions options)
 
             _configuration = await LoadConfigurationAsync(_options.Config);
 
+            CreateThumbprint();
```
joperezr, line 173: NIT: is it worth it to add a comment here explaining why we need to make sure this is called here for future reference?

joperezr (approved):
Small NIT but looks good otherwise.

PR #581, opened 2023-04-27
Update Octokit version

sebastienros:


joperezr (approved):
LGTM. As an aside, have we considered setting up dependabot for these to get automatic dependency update PRs?

sebastienros:
I prefer to handle updates when necessary. Compliance will flag any unsecure version. That reduces the maintenance work to do: approving regular PRs even though the product doesn't need them.

PR #584, opened 2023-05-05
Add options to support grpc requests

sebastienros:
- Method argument
- Base64 encoded body
- Content headers

### DIFF
--- a/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
+++ b/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
@@ -41,13 +41,14 @@ namespace Microsoft.Crank.Jobs.HttpClientClient
         public static int ExecutionTimeSeconds { get; set; }
         public static int Connections { get; set; }
         public static List<string> Headers { get; set; }
-        public static string Body { get; set; }
+        public static byte[] Body { get; set; }
         public static Version Version { get; set; }
         public static string CertPath { get; set; }
         public static string CertPassword { get; set; }
         public static X509Certificate2 Certificate { get; set; }
         public static bool Quiet { get; set; }
         public static string Format { get; set; }
+        public static HttpMethod Method { get; set; } = HttpMethod.Get;
         public static bool Local { get; set; }
         public static Timeline[] Timelines {  get; set; }
         public static string Script { get; set; }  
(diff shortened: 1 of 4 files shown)


BrennanConroy (approved):

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -477,28 +514,41 @@ public static async Task<WorkerResult> DoWorkAsync()
...
                 foreach (var header in Headers)
                 {
-                    var headerNameValue = header.Split(" ", 2, StringSplitOptions.RemoveEmptyEntries);
-                    requestMessage.Headers.Remove(headerNameValue[0]);
-                    requestMessage.Headers.TryAddWithoutValidation(headerNameValue[0], headerNameValue[1]);
+                    if (header.Split(':', 2, StringSplitOptions.RemoveEmptyEntries) is [var key, var value])
+                    {
+                        if (!requestMessage.Headers.TryAddWithoutValidation(key, value.TrimStart()))
+                        {
+                            requestMessage.Content.Headers.TryAddWithoutValidation(key, value.TrimStart());
```
JamesNK, line 541: If there is no body then `Content` is null and this will error.

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -477,28 +514,41 @@ public static async Task<WorkerResult> DoWorkAsync()
...
+                {
+                    requestMessage.Content = new ByteArrayContent(Body);
+                }
+
                 foreach (var header in timeline.Headers)
                 {
-                    requestMessage.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                    if (!requestMessage.Headers.TryAddWithoutValidation(header.Key, header.Value.TrimStart()))
+                    {
+                        requestMessage.Content.Headers.TryAddWithoutValidation(header.Key, header.Value.TrimStart());
```
JamesNK, line 530: If there is no body then `Content` is null and this will error.

sebastienros: Thanks, will provide a PR to fix it

PR #585, opened 2023-05-08
Ignore content headers if there is no content

sebastienros:


JamesNK (approved):
Is it worth throwing an error if the header is never added with TryAdd to either collection?

Or logging a message?

sebastienros:
I believe that if the header is not added then the body is missing, so probably this should be a better signal something is missing. Also almost nothing is using this feature. Logging would be bad since this code runs for each request. An exception might be too harsh. I'll change my mind when someone complains.

PR #592, opened 2023-06-02
Remove explicit tfms from clients

sebastienros:
Will default to the csproj's one.


joperezr (approved):
LGTM. Just curious, what is the default?

sebastienros:
The default is to align with the value from the csproj file. So removing is not only fixing the current bug (doesn't compile with this sdk as it's using new c# constructs) it will also follow the tfm when we upgrade the projects.

sebastienros:
The other most commonly used clients (wrk, bombardier, h2load) were already configured this way. It happens that I recently updated HttpClient and now some benmchmarks started to fail. I was using array deconstruction.

PR #602, opened 2023-07-25
Pin latest SDK to unblock benchmarks on net8.0

sebastienros:


eerhardt:
The latest SDK builds work now. Closing.

PR #603, opened 2023-07-28
Improve first request measurement resiliency

sebastienros:
Fixes slow benchmarks that fail on the first request

### DIFF
--- a/src/Microsoft.Crank.Jobs.Bombardier/Program.cs
+++ b/src/Microsoft.Crank.Jobs.Bombardier/Program.cs
@@ -364,8 +364,7 @@ namespace Microsoft.Crank.Jobs.Bombardier
             var cts = new CancellationTokenSource(30000);
             var httpMessage = new HttpRequestMessage(HttpMethod.Get, url);
 
-            var stopwatch = new Stopwatch();
-            stopwatch.Start();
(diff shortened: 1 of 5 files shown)


src/Microsoft.Crank.Jobs.Bombardier/Program.cs
```diff
@@ -382,6 +381,15 @@ public static async Task MeasureFirstRequest(string[] args)
             {
                 Console.WriteLine("A timeout occurred while measuring the first request");
             }
+            catch (HttpRequestException)
+            {
+                Console.WriteLine("A connection exception occurred while measuring the first request");
```
BrennanConroy, line 386: Any useful info on HttpRequestException? `Message`?

sebastienros: This is the exceptions hierarchy; the message is not really useful per se

```
[STDERR] Unhandled exception. System.Net.Http.HttpRequestException: An error occurred while sending the request.
[STDERR]  ---> System.IO.IOException: Unable to read data from the transport connection: Connection reset by peer.
[STDERR]  ---> System.Net.Sockets.SocketException (104): Connection reset by peer
```

BrennanConroy (approved):

PR #604, opened 2023-07-28
Cache records endpoint across message and skip failures due to tooling

sebastienros:


src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -34,10 +36,10 @@ public enum ResultTypes
         public string TaskInstanceId { get; set; }
         public string AuthToken { get; set; }
 
-        public Records Records { get; set; }
-        private DateTime _lastRecordsRefresh = DateTime.UtcNow;
         private static readonly JsonSerializerOptions _serializationOptions = new JsonSerializerOptions { PropertyNameCaseInsensitive = true };
 
+        private MemoryCache _memoryCache = new MemoryCache(new MemoryCacheOptions());
```
sebastienros, line 40: Cache API result across Service Bus messages (benchmarks) since these are for the same pipeline.

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -254,28 +256,30 @@ public async Task<bool> UpdateTaskTimelineRecordAsync(string taskLogObject)
...
-
-                var result = await GetDataAsync(getRecordsUrl);
+                // The application is single-threaded
 
-                if (result.IsSuccessStatusCode)
+                var records = await _memoryCache.GetOrCreateAsync(getRecordsUrl, async entry =>
                 {
+                    var result = await GetDataAsync(getRecordsUrl);
+
+                    result.EnsureSuccessStatusCode();
```
sebastienros, line 270: Any failure is bubble up and the benchmark will be put back in the queue

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -107,10 +107,12 @@ private static async Task MessageHandler(ProcessMessageEventArgs args)
...
                 {
-                    Console.ForegroundColor = ConsoleColor.DarkRed;
-                    Console.WriteLine($"{LogNow} Could not retrieve records...");
-                    Console.ResetColor();
+                    await devopsMessage?.SendTaskCompletedEventAsync(DevopsMessage.ResultTypes.Skipped);
 
+                    Console.WriteLine($"{LogNow} Could not retrieve Records, skipping...");
+
+                    // Release the message for further processing
+                    await args.AbandonMessageAsync(message);
```
sebastienros, line 115: REST API failure, abandon the message such that the next loop can try it again

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -247,7 +252,7 @@ record = records.Value.FirstOrDefault(x => x.Id == devopsMessage.TaskInstanceId)
 
                 try
                 {
-                    await devopsMessage?.SendTaskCompletedEventAsync(DevopsMessage.ResultTypes.Failed);
+                    await devopsMessage?.SendTaskCompletedEventAsync(DevopsMessage.ResultTypes.SucceededWithIssues);
```
sebastienros, line 255: There was an exception during message processing, mark the benchmark in orange.

src/Microsoft.Crank.AzureDevOpsWorker/DevopsMessage.cs
```diff
@@ -254,28 +255,30 @@ public async Task<bool> UpdateTaskTimelineRecordAsync(string taskLogObject)
...
                 {
+                    var result = await GetDataAsync(getRecordsUrl);
+
+                    result.EnsureSuccessStatusCode(); 
+                    
                     var content = await result.Content.ReadAsStringAsync();
 
-                    Records = JsonSerializer.Deserialize<Records>(content, _serializationOptions);
-                }
+                    var records = JsonSerializer.Deserialize<Records>(content, _serializationOptions); ;
```
BrennanConroy, line 274:
```suggestion
                    var records = JsonSerializer.Deserialize<Records>(content, _serializationOptions);
```

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
BrennanConroy: We're abandoning so that service bus will retry it, but telling AzDo that the job is skipped?

sebastienros: Yes, because we told AzDo to "start" it so the blue icon is spinning (it's fake it's a server-less job) so to stop it we make it skipped, then the next time it's taken it will be marked running again (spinning blue) and eventually fail (red) or (success with issues, orange)

BrennanConroy: I guess I don't understand the flow here. Why does AzDo run it again?

sebastienros: AzDo runs a PublishServiceBus task, which only generates a message in the queue and doesn't update the task itself (white circle). Then as we read them the message contains a callback url to update the task in azdo with state (completed, status, log, ...). As we run the benchmark we mark it "running", then "completed" with a status (failed, success, warning).

BrennanConroy: Ok, so here we update it with skipped. But then we reprocess the message because we abandoned it and service bus resends it? Can you then update a previously completed AzDo task?

sebastienros: I think so, the REST endpoint is the same for start/complete(+status): a POST to `{PlanUrl}/{ProjectId}/_apis/distributedtask/hubs/{HubName}/plans/{PlanId}/events`. There is nothing that tells you can't change from one state to another (and there are other ones than these two)

If only I could find some doc :/

BrennanConroy (approved):
Yolo

PR #606, opened 2023-08-01
Add http requests retry logic to http AzureDevOpsWorker

sebastienros:


src/Microsoft.Crank.AzureDevOpsWorker/RetryHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+        { }
+
+        protected override async Task<HttpResponseMessage> SendAsync(
+            HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
+            HttpResponseMessage response = null;
+            for (int i = 0; i < MaxRetries; i++)
+            {
+                response = await base.SendAsync(request, cancellationToken);
```
BrennanConroy, line 26: Should there be any delay and/or logging between attempts?

sebastienros: Thanks, done

BrennanConroy: `HttpResponseMessage` should probably be disposed if it's not being returned?

BrennanConroy (approved):

src/Microsoft.Crank.AzureDevOpsWorker/RetryHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Crank.AzureDevOpsWorker
+{
+    internal class RetryHandler : DelegatingHandler
```
eerhardt, line 12: Can/should we use https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience?

sebastienros: 🤟

PR #607, opened 2023-08-01
Add job retries support to AzureDevOpsWorker

sebastienros:


src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -160,57 +160,70 @@ private static async Task MessageHandler(ProcessMessageEventArgs args)
...
-                            Console.WriteLine($"{LogNow} Job is completed ({record.Result}), interrupting...");
+                            if (record != null && record?.State == "completed")
+                            {
+                                Console.WriteLine($"{LogNow} Job is completed ({record.Result}), interrupting...");
+
+                                driverJob.Stop();
+                            }
 
-                            driverJob.Stop();
+                            await Task.Delay(TaskLogFeedDelay);
```
BrennanConroy, line 221: Should this be in an `else`? No need to delay if we stopped the driver right?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -160,57 +160,70 @@ private static async Task MessageHandler(ProcessMessageEventArgs args)
...
-                                driverJob.Stop();
-                            }                                
-                        }
+                                if (!success)
+                                {
+                                    Console.ForegroundColor = ConsoleColor.DarkYellow;
+                                    Console.WriteLine($"{LogNow} SendTaskLogFeedsAsync failed. If the task was canceled, this jobs should be stopped.");
+                                    Console.ResetColor();
+
+                                    driverJob.Stop();
```
BrennanConroy, line 202: `continue;`?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
sebastienros: Better

BrennanConroy (approved):

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
sebastienros: I see the point, so I removed this line and let the next part do it. But I used this in the timeout detection instead

PR #610, opened 2023-08-07
Update dotnet versions

sebastienros:
Attempting to fix this signing validation error:

```
D:\a\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\7.0.0-beta.23370.3\tools\SdkTasks\SigningValidation.proj : error : Version 8.0.100-preview.6.23330.14 of the .NET SDK requires at least version 17.6.0 of MSBuild. The current available version of MSBuild is 17.4.0.51802. Change the .NET SDK specified in global.json to an older version that requires the MSBuild version currently available.
D:\a\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\7.0.0-beta.23370.3\tools\SdkTasks\SigningValidation.proj : error : The NuGetSdkResolver did not resolve this SDK because there was no version specified in the project or global.json.
D:\a\_work\1\s\.packages\microsoft.dotnet.arcade.sdk\7.0.0-beta.23370.3\tools\SdkTasks\SigningValidation.proj : error MSB4236: The SDK 'Microsoft.NET.Sdk' specified could not be found.

```

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,6 @@
 {
   "tools": {
     "dotnet": "7.0.306"
-    }
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "7.0.0-beta.23370.3"
--- a/global.json
+++ b/global.json
@@ -1,13 +1,6 @@
 {
   "tools": {
-    "dotnet": "7.0.306",
-    "runtimes": {
-      "dotnet": [
-        "7.0.9"
-      ],
-      "aspnetcore": [
-        "7.0.9"
-      ]
+    "dotnet": "7.0.306"
     }
   },
   "msbuild-sdks": {
--- a/global.json
+++ b/global.json
@@ -1,12 +1,12 @@
 {
   "tools": {
-    "dotnet": "8.0.100-preview.6.23330.14",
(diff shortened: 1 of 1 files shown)


global.json
```diff
@@ -1,12 +1,12 @@
 {
   "tools": {
-    "dotnet": "8.0.100-preview.6.23330.14",
+    "dotnet": "7.0.306",
     "runtimes": {
       "dotnet": [
-        "7.0.1"
+        "7.0.9"
```
joperezr, line 6: NIT: You no longer need the runtimes section. It was only required before because we were using a .NET 8 SDK which only comes with .NET 8 sharedfx, so we needed to additionally restored 7.0. Now that we are using 7.0 SDK, sharedFX should be included so no need for that section.

joperezr (approved):
LGTM, apart from the comment that I shared.

global.json
sebastienros: Should I also set `sdk: ...` ?

joperezr: You can if you want. The tools section is mostly just so that our tooling will restore it automatically, the sdk property is so that people that don't have this version installed won't be able to call dotnet commands in the repo (essentially you force people to have that sdk or higher.)

PR #611, opened 2023-08-07
Fix potential NRE when yaml file is incomplete

sebastienros:
I believe the fact that the YAML doesn't contain labels/owners here could be the reason of an NRE.
https://github.com/aspnet/Benchmarks/blob/7b44db144e31b4fd0fc648315713ada14a09ccb5/build/regressions.config.yml#L134

### DIFF
--- a/src/Microsoft.Crank.RegressionBot/Program.cs
+++ b/src/Microsoft.Crank.RegressionBot/Program.cs
@@ -161,10 +161,13 @@ namespace Microsoft.Crank.RegressionBot
                 {
                     var configuration = await LoadConfigurationAsync(configurationFilenameOrUrl);
                     sources.AddRange(configuration.Sources);
-                    
-                    foreach (var template in configuration.Templates)
+
+                    if (configuration.Templates != null)
                     {
-                        templates[template.Key] = template.Value;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -161,10 +161,13 @@ private static async Task<int> Controller(BotOptions options)
                 {
                     var configuration = await LoadConfigurationAsync(configurationFilenameOrUrl);
                     sources.AddRange(configuration.Sources);
-                    
-                    foreach (var template in configuration.Templates)
+
+                    if (configuration.Templates != null)
                     {
-                        templates[template.Key] = template.Value;
+                        foreach (var template in configuration.Templates)
```
BrennanConroy, line 167: Why doesn't this one get the same `?? Enumerable.Empty` treatment?

sebastienros: It's a `Dictionary<string,string>`, `IEnumerable<KeyValuePair<string,string>>` might be too ugly

BrennanConroy (approved):

PR #617, opened 2023-08-11
Logging when publish is skipped

sebastienros:


sebastienros:
I am trying to understand why the results are not published. I am already catching the exception if the creation of the comment is failing, so last reason would be for this flag to be false, though with my local tests the command line parsing works correctly.

adityamandaleeka (approved):

captainsafia (approved):
LGTM! Would it be helpful to include the issue ID or other references in the log message?

PR #627, opened 2023-08-21
Add wrk for arm64 support

sebastienros:


LoopedBard3 (approved):
LGTM!

src/Microsoft.Crank.Jobs.Wrk/WrkProcess.cs
```diff
@@ -18,7 +19,8 @@ namespace Microsoft.Crank.Wrk
     static class WrkProcess
     {
         private static string _wrkFilename;
-        const string WrkUrl = "https://aspnetbenchmarks.blob.core.windows.net/tools/wrk";
+        const string WrkLinuxAmd64 = "https://aspnetbenchmarks.blob.core.windows.net/tools/wrk-linux-amd64";
+        const string WrkLinuxArm64 =  "https://aspnetbenchmarks.blob.core.windows.net/tools/wrk-linux-arm64";
```
RobertHenry6bev, line 23: I'd like to see the code that is run, somewhere, to checkout and build these binaries.  techempower results depend on the throughput of wrk.

sebastienros: @RobertHenry6bev I use a docker image to build these, and then extract the binary to azure blobs since the github repository doesn't publish them like bombardier does.
Here is the script that I use:
https://github.com/William-Yeh/docker-wrk/blob/master/Dockerfile

sebastienros: And you can try to download these files and compare them to your local build with a load test

PR #629, opened 2023-08-21
Increase stored hash to prevent collisions

sebastienros:
Small hashes in urls may collide and GH can't generate the diff in these cases.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -68,6 +68,7 @@ namespace Microsoft.Crank.Agent
 
         private static readonly string DefaultTargetFramework = "net6.0";
         private static readonly string DefaultChannel = "current";
+        private const int CommitHashLength = 12;
 
         private const string PerfViewVersion = "v3.0.7";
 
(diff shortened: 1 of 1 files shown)


sebastienros:
Use 12 instead of 7, tried the latest commits we used and 8 would have been sufficient.

sebastienros:
If you want to try, here is an example `https://github.com/dotnet/runtime/compare/ff9beed32cb2c86ccdd14892483b0fcc831cc5d6...e89794659669cb7bb967db73a7ea6889c3891727`

Try to reduce the size of the hashes

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3083,7 +3083,7 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
                     {
                         Name = Measurements.BenchmarksAspNetCoreVersion,
                         Timestamp = DateTime.UtcNow,
-                        Value = $"{aspNetCoreVersion}+{aspnetCoreCommitHash.Substring(0, 7)}"
+                        Value = $"{aspNetCoreVersion}+{aspnetCoreCommitHash.Substring(0, 12)}"
```
eerhardt, line 3086: (nit) do you want to make a const field for this, so it is only in 1 spot?

eerhardt (approved):

PR #631, opened 2023-08-22
Update msquic to latest published package

CarnaViire:
Installing latest package from packages.microsoft.com (.NET 8 bumped required version to 2.2.2+).

We dropped QUIC support from .NET 6, so I removed msquic 1.9.

Fixes https://github.com/aspnet/Benchmarks/issues/1894

cc @sebastienros

### DIFF
--- a/docker/agent/Dockerfile
+++ b/docker/agent/Dockerfile
@@ -39,31 +39,13 @@ RUN apt-get update \
         libkrb5-dev
 
 # Install HTTP/3 support
-
-# msquic 2+ for .NET 7+
-ENV MSQUIC_VERSION 2.2.1
-RUN test "$(uname -m)" != 'aarch64' && \
-    cd /tmp && \
-    curl -O "https://packages.microsoft.com/debian/10/prod/pool/main/libm/libmsquic/libmsquic_${MSQUIC_VERSION}_amd64.deb" && \
-    dpkg -i "libmsquic_${MSQUIC_VERSION}_amd64.deb" && \
-    rm "libmsquic_${MSQUIC_VERSION}_amd64.deb" || true
-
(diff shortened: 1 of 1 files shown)


docker/agent/Dockerfile
```diff
@@ -39,31 +39,13 @@ RUN apt-get update \
...
-    mkdir libmsquic_1.9 && \
-    dpkg-deb -x libmsquic_1.9.1_amd64.deb libmsquic_1.9 && \
-    cp libmsquic_1.9/usr/lib/x86_64-linux-gnu/libmsquic.so /usr/lib/x86_64-linux-gnu/ && \
-    cp libmsquic_1.9/usr/lib/x86_64-linux-gnu/libmsquic.lttng.so /usr/lib/x86_64-linux-gnu/ && \
-    rm -rf libmsquic_1.9* || true
-
+RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
+    echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc| sha256sum --check - && \
+    apt-key add microsoft.asc && \
+    apt-add-repository https://packages.microsoft.com/debian/11/prod && \
```
sebastienros, line 45: Does it not matter that it's pointing to Debian/11 while the base image here (sdk:7) is based on Debian/10?

CarnaViire: sdk:7.0 is based on Debian 11 according to https://hub.docker.com/_/microsoft-dotnet-sdk/
image

sebastienros: Oh, thanks, I thought the latest was 11. I had to update to 8-preview recently and now I see that I can actually stay on 7 and use a debian/12.

PR #634, opened 2023-08-29
Add support for net9.0

sebastienros:
Downloads SDK and runtimes from 9.0 feeds when `--application.framework net9.0` is used. The TFM is the csproj is kept to net8.0 since net9.0 doesn't exist already. But net9.0 is detected and uses the new feeds.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Crank.Agent
         private static readonly string DefaultChannel = "current";
         private const int CommitHashLength = 12;
 
-        private const string PerfViewVersion = "v3.0.7";
+        private const string PerfViewVersion = "v3.1.5";
 
         private static readonly HttpClient _httpClient;
         private static readonly HttpClientHandler _httpClientHandler;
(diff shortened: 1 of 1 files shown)


sebastienros:
Local run results:

```
| application             |                                     |
| ----------------------- | ----------------------------------- |
| Max CPU Usage (%)       | 46                                  |
| Max Cores usage (%)     | 743                                 |
| Max Working Set (MB)    | 69                                  |
| Max Private Memory (MB) | 69                                  |
| Build Time (ms)         | 27,369                              |
| Start Time (ms)         | 2,092                               |
| Published Size (KB)     | 102,390                             |
| Symbols Size (KB)       | 53                                  |
| .NET Core SDK Version   | 9.0.100-alpha.1.23429.3             |
| ASP.NET Core Version    | 9.0.0-alpha.1.23428.13+9402bfac90a6 |
| .NET Runtime Version    | 9.0.0-alpha.1.23429.6+dff23ec24c41  |
```

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2766,19 +2779,19 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
...
-            sdkVersion = await ResolveSdkVersion(sdkVersion, currentSdkVersion);
+            sdkVersion = await ResolveSdkVersion(sdkVersion, currentSdkVersion, targetFramework);
 
             aspNetCoreVersion = await ResolveAspNetCoreVersion(aspNetCoreVersion, currentAspNetCoreVersion, targetFramework);
 
             sdkVersion = PatchOrCreateGlobalJson(job, benchmarkedApp, sdkVersion);
 
             var installAspNetSharedFramework = job.UseRuntimeStore
-                || aspNetCoreVersion.StartsWith("3.0")
                 || aspNetCoreVersion.StartsWith("3.1")
```
BrennanConroy, line 2789: Remove 3.1 and 5.0 as well?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4297,9 +4311,9 @@ private static async Task<string> GetAspNetRuntimeVersion(string buildToolsPath,
         /// </summary>
         private static async Task<(string Runtime, string Desktop, string AspNet, string Sdk)> GetCurrentVersions(string targetFramework)
         {
-            // There are currently no release for net8.0
+            // There are currently no release for net9.0
             // Remove once there is at least a preview and a "release-metadata" file
```
BrennanConroy, line 4315: 😆 we never removed it 8.0, does this actually do anything then?

src/Microsoft.Crank.Agent/Startup.cs
sebastienros: I removed it in other parts so I assume it's logical to do it everywhere, will scan the file for these, thanks.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -66,7 +66,7 @@ public class Startup
             Based on the target framework
          */
 
-        private static readonly string DefaultTargetFramework = "net6.0";
+        private static readonly string DefaultTargetFramework = "net7.0";
```
eerhardt, line 69: Really? `net6.0` is going to be supported longer than 7. Do we just want to jump to 8?

eerhardt (approved):
Seems fine

src/Microsoft.Crank.Agent/Startup.cs
sebastienros: Yes, right no we run all the benchmarks with the latest versions of 8.0, but if we force the "current" channel (`--application.channel current`) then it would for the latest public release (preview-7 right now).

src/Microsoft.Crank.Agent/Startup.cs
sebastienros: It's just if we don't set any framework it will use `net7.0`, once `net8.0` is released officially we can switch. Note that it is never used in our benchmarks since we always set it explicitly.

PR #638, opened 2023-09-07
Fix NoBuild support

caaavik-msft:
Still doing some testing on this, but from initial testing this seems to work and builds will be reused as long as all the build inputs (see BuildKeyData) are identical.


sebastienros:
So each build cache actually has its own source cache, and the build folder still local to the source ( in ./published)? That would be fine.

sebastienros (approved):

PR #640, opened 2023-09-22
Delete NOTICE.txt

sebastienros:
Q: Does my public GitHub repository need a NOTICE?
A: Yes, if you have copied anything that is under an open source license into your repo, then your repo needs a NOTICE. If you only reference open source sitting outside your repo (using a package manager or git submodules), then that referenced open source does not need to be included on a NOTICE. For instructions on creating a NOTICE for your repo, see NOTICE in a Microsoft open source project.


sebastienros:
I actually believe it's required for other reasons than the one mentioned here.

PR #646, opened 2023-10-19
Update SQL client version

javiercn:


sebastienros (approved):

javiercn:
@sebastienros we might need to do a bit more. The underlying culprit is the Azure.Identity dependency

PR #648, opened 2023-10-19
Remove cross-org build step

sebastienros:
Matching PR: https://github.com/dotnet/arcade/pull/11577

### DIFF
--- a/eng/common/templates/post-build/common-variables.yml
+++ b/eng/common/templates/post-build/common-variables.yml
@@ -1,8 +1,4 @@
 variables:
-  - group: AzureDevOps-Artifact-Feeds-Pats
-  - group: DotNet-Blob-Feed
-  - group: DotNet-DotNetCli-Storage
-  - group: DotNet-MSRC-Storage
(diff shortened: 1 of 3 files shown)


sebastienros:
Signing validation doesn't work anymore with this step since the variable group was removed

eng/common/templates/post-build/common-variables.yml
```diff
@@ -1,8 +1,4 @@
 variables:
-  - group: AzureDevOps-Artifact-Feeds-Pats
```
sebastienros, line 2: An arcade update re-added, this is wrong

BrennanConroy (commented):
Update publish-build-assets.yml too? Should this repo get arcade updates automatically?

BrennanConroy (approved):

sebastienros:
> Should this repo get arcade updates automatically?

It does, and that's the issue. It syncs with release/7.0 which still has these variables defined, but they are removed from ADO. So the build would fail. I had removed it before but then a sync added it back and I didn't pay attention.

PR #653, opened 2023-10-26
Remove some ENVs

sebastienros:
Remove some hard coded environment variables which should be set in config files instead.


BrennanConroy:
Remove https://github.com/dotnet/crank/blob/da5b39bef6e229f63ac061062f0e66db293d8c65/src/Microsoft.Crank.Agent/Startup.cs#L4450C12-L4450C65?
This won't break anyone right?

BrennanConroy (approved):

PR #654, opened 2023-11-03
Apply Job Object in delayed child process

sebastienros:


sebastienros:
I filed an issue in the runtime https://github.com/dotnet/runtime/issues/94364 that would explain why the `ProcessorCount` is wrong with CPU affinity on windows. I believed that is was a timing issue, but even as I fixed that it was still not reporting the correct value, whereas Memory limits were correctly read. I am keeping the changes in crank to use a wrapper application that waits for the JobObject to be applied before starting the benchmarked application until this is fixed, and also this one https://github.com/dotnet/runtime/issues/94127
 
What I noticed also is that `cpuLimitRatio` is working as expected in Crank, like memory limits, since both are on the same job object. The cpu affinity is not on a job object actually but using a custom PInvoke call. So maybe setting `cpuRatio` would work just fine instead of cpu affinity, and we would just need to grab the corresponding number of CPUs in the stats.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4643,6 +4644,33 @@ private static async Task<Process> StartProcess(string hostname, string benchmar
                 process.WaitForExit();
             };
 
+            var useWindowsLimiter = (job.MemoryLimitInBytes > 0 || job.CpuLimitRatio > 0 || !String.IsNullOrWhiteSpace(job.CpuSet)) && RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
joperezr, line 4647: Nit: not that it really matters, but you could do the PS check first which would save you from running the rest of the statements in case it's not Windows.

src/Microsoft.Crank.JobOjectWrapper/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+    Console.WriteLine("Arguments are missing.");
+    Environment.Exit(-1);
+}
+
+Console.WriteLine("Job Object Wrapper");
+Console.WriteLine("Waiting for Job Object to be setup...");
+
+await Task.Delay(1000);
+
+var process = new Process()
```
joperezr, line 18: FYI Arcade has this thing called RemoteExecutor which essentially does the same here I think. No meed to change, just in case you want to consider.

joperezr (approved):
LGTM

src/Microsoft.Crank.Agent/Startup.cs
teo-tsirpanis, line 4647:
```suggestion
            var useWindowsLimiter = (job.MemoryLimitInBytes > 0 || job.CpuLimitRatio > 0 || !String.IsNullOrWhiteSpace(job.CpuSet)) && OperatingSystem.IsWindows();
```

sebastienros: Is there any description of what RemoteExecutor's goal is exactly. I didn't get the relationship with this after a quick look at the code.

src/Microsoft.Crank.JobOjectWrapper/Program.cs
sebastienros: Is there any description of what RemoteExecutor's goal is exactly. I didn't get the relationship with this after a quick look at the code.

PR #655, opened 2023-11-06
Keep controller connection alive while uploading files

sebastienros:
```
[02:43:55.502] Uploading C:\Users\mirogers.REDMOND\AppData\Local\Temp\tmpB997.tmp (334,718KB)
[02:44:20.668] Could not ping the agent on 'asp-perf-lin:5001', retrying ...
[02:44:25.807] Could not ping the agent on 'asp-perf-lin:5001', retrying ...
[02:44:30.931] Could not ping the agent on 'asp-perf-lin:5001', retrying ...
Unhandled exception. System.InvalidOperationException: An error occurred while uploading a file.
---> System.Net.Http.HttpRequestException: Error while copying content to a stream.
---> System.IO.IOException: Unable to write data to the transport connection: An existing connection was forcibly closed by the remote host..
```

### DIFF
--- a/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
+++ b/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
@@ -325,7 +325,15 @@ namespace Microsoft.Crank.Agent.Controllers
 
             var tempFilename = Path.GetTempFileName();
 
-            await SaveBodyAsync(job, tempFilename);
+            try
+            {
+                job.LastDriverCommunicationUtc = DateTime.UtcNow;
+                await SaveBodyAsync(job, tempFilename);
+            }
+            catch
+            {
+                return StatusCode(500, "Error while saving file");
+            }
 
             job.Attachments.Add(new Attachment
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -454,20 +454,31 @@ public async Task<IActionResult> UploadBuildFile(int id)
...
+                task = Request.Body.CopyToAsync(outputFileStream, Request.HttpContext.RequestAborted);
+            }
+
+            // Keeps controller's connection alive as PINGs might not be accepted
+            // by the agent while big files are uploaded.
+
+            while (!Request.HttpContext.RequestAborted.IsCancellationRequested && !task.IsCompleted)
+            {
+                await Task.Delay(500);
+                job.LastDriverCommunicationUtc = DateTime.UtcNow;
```
BrennanConroy, line 481: This is the important part I assume?

sebastienros: It is, is that wrong ;)

BrennanConroy: Just checking my understanding

BrennanConroy: Should we await the task if it completes in case there is an exception?

BrennanConroy (approved):

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -367,15 +375,18 @@ public async Task<IActionResult> UploadAttachmentZip(int id)
...
 
-            await SaveBodyAsync(tempFilename);
-
-            job.LastDriverCommunicationUtc = DateTime.UtcNow;
-
             try
             {
+                job.LastDriverCommunicationUtc = DateTime.UtcNow;
+                await SaveBodyAsync(job, tempFilename);
+
```
mikelle-rogers, line 382: Why did you change the order of these two lines?

sebastienros: Doesn't matter much, this will be updated every second by the loop which is more than the 10s timeout.

PR #658, opened 2023-11-07
Call Microsoft.Crank.JobObjectWrapper.exe with full path 

LoopedBard3:
Instead of relying on starting in the correct working directory, specify that we want the Microsoft.Crank.JobObjectWrapper.exe in the same location of the assembly/tool.

### DIFF
--- a/src/Microsoft.Crank.JobOjectWrapper/Program.cs
+++ b/src/Microsoft.Crank.JobOjectWrapper/Program.cs
@@ -30,8 +30,26 @@ Console.WriteLine("Starting process...");
 Console.WriteLine($"Filename: {process.StartInfo.FileName}");
 Console.WriteLine($"Args: {process.StartInfo.Arguments}");
 
+process.OutputDataReceived += (_, e) =>
+{
+    if (e != null && e.Data != null)
+    {
+        Console.WriteLine(e.Data);
+    }
+};
+
+process.ErrorDataReceived += (_, e) =>
+{
+    if (e != null && e.Data != null)
+    {
+        Console.WriteLine(e.Data);
+    }
+};
+
 process.Start();
 Console.WriteLine($"##ChildProcessId:{process.Id}");
+process.BeginOutputReadLine();
+process.BeginErrorReadLine();
 process.WaitForExit();
 
 Console.WriteLine(process.StandardOutput.ReadToEnd());
--- a/src/Microsoft.Crank.JobOjectWrapper/Program.cs
+++ b/src/Microsoft.Crank.JobOjectWrapper/Program.cs
@@ -20,9 +20,7 @@ var process = new Process()
     StartInfo = {
         FileName = args[0],
         Arguments = string.Join(" ", args[1..]),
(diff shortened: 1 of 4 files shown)


sebastienros:
Can you add a functional test that will set cpuset to `0` just to trigger this code path? By copying exactly one of the existing one?

sebastienros:
Example of one that can be used" https://github.com/dotnet/crank/blob/main/test/Microsoft.Crank.IntegrationTests/CommonTests.cs#L56

src/Microsoft.Crank.JobOjectWrapper/Program.cs
```diff
@@ -30,8 +30,26 @@
 Console.WriteLine($"Filename: {process.StartInfo.FileName}");
 Console.WriteLine($"Args: {process.StartInfo.Arguments}");
 
+process.OutputDataReceived += (_, e) =>
```
sebastienros, line 33: I think it's equivalent to not invoking these:
```
RedirectStandardOutput = true,
RedirectStandardError = true,
```

LoopedBard3: These are both already being set in the StartInfo of the process starting this wrapper and the process being started by this wrapper.

LoopedBard3:
> Can you add a functional test that will set cpuset to `0` just to trigger this code path? By copying exactly one of the existing one?

I added a test that uses --application.cpuSet 0 to test the code path. It would probably be good to also add one for a cpuSet range and one for multiple cpuSet ranges. Although I don't know the processor count of the normal test agents.

src/Microsoft.Crank.JobOjectWrapper/Program.cs
```diff
@@ -30,8 +30,26 @@
...
+    }
+};
+
 process.Start();
 Console.WriteLine($"##ChildProcessId:{process.Id}");
+process.BeginOutputReadLine();
+process.BeginErrorReadLine();
 process.WaitForExit();
 
 Console.WriteLine(process.StandardOutput.ReadToEnd());
```
LoopedBard3, line 55: Is this line still necessary with this setup?

src/Microsoft.Crank.JobOjectWrapper/Program.cs
LoopedBard3: As far as I can tell, just having the redirect enabled does not cause the parent process to actually receive the output live, and we need the output live in order to tell when the application has started.

sebastienros: > As far as I can tell, just having the redirect enabled does not cause the parent process to actually receive the output live

Was wondering the opposite, remove the code completely.

sebastienros (approved):

src/Microsoft.Crank.JobOjectWrapper/Program.cs
LoopedBard3: Looks like that works as well, went ahead and updated to remove code rather than add code.

PR #661, opened 2023-11-21
Detect perfview failures

sebastienros:
This uses `/LogFile:` which is the documented way to run perfview without user interaction. If the process fails to start then the log file content is returned to the user. Tested with wrong perfview arguments:

```
[05:57:41.252] 'application' is now building ... http://localhost:5010/jobs/1/buildlog
[05:58:00.451] 'application' failed on agent, stopping...
PerfView logging started at 11/20/2023 5:57:59 PM
Error: The value 'GCHeapAndTypeName' for parameter ClrEvents is not a member of the enumeration Keywords.
The legal values are either a decimal integer, 0x and a hex integer or
    None
    GC
    GCHandle
    Binder
    Loader
    Jit
    NGen
    StartEnumeration
    StopEnumeration
    Security
...
```

/cc @Maoni0 @markples

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Crank.Agent
         private static readonly string DefaultChannel = "current";
         private const int CommitHashLength = 12;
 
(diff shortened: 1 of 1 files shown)


Maoni0 (approved):
thanks!

PR #669, opened 2023-12-13
Inject updated list of feeds in csproj files

sebastienros:
Use `<RestoreAdditionalProjectSources/>` in csproj to link to nightly build feeds since the root NuGet.config might be invalidated by local NuGet.config `<clear />` statements.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -41,7 +41,6 @@ using Repository;
 using Serilog;
 using Serilog.Events;
 using Serilog.Sinks.SystemConsole.Themes;
-using Vanara.PInvoke;
 using OperatingSystem = Microsoft.Crank.Models.OperatingSystem;
 
 namespace Microsoft.Crank.Agent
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -4620,8 +4616,9 @@ private static async Task<Process> StartProcess(string hostname, string benchmar
                 var arguments = process.StartInfo.Arguments;
 
                 // The executable should be in the same folder as the agent since it references the Console project
-                process.StartInfo.FileName = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Microsoft.Crank.JobObjectWrapper.exe");
-                process.StartInfo.Arguments = filename + " " + arguments;
+                // Use 'dotnet exe .dll' to use the current default dotnet version or the tests could fail if the .exe doesn't match what version is available locally
```
caaavik-msft, line 4619:
```suggestion
                // Use 'dotnet exec .dll' to use the current default dotnet version or the tests could fail if the .exe doesn't match what version is available locally
```
Noticed a small type here. I didn't realise that `dotnet exec .dll` was a thing until now, It seems that it is equivalent to just doing `dotnet .dll` but I guess it doesn't hurt to use `exec`

caaavik-msft (approved):

PR #673, opened 2023-12-15
Update Dockerfile for .net8.0

sebastienros:
libmsquic updated commands
python3 tools

### DIFF
--- a/docker/agent/Dockerfile
+++ b/docker/agent/Dockerfile
@@ -40,14 +40,14 @@ RUN apt-get update \
         # .NET 9.0 requirement
         libc6 
 
-# Install HTTP/3 support
 RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc| sha256sum --check - && \
     apt-key add microsoft.asc && \
-    apt-add-repository https://packages.microsoft.com/debian/11/prod && \
+    rm microsoft.asc && \
+    echo deb https://packages.microsoft.com/debian/12/prod bookworm main >> /etc/apt/sources.list.d/microsoft.list && \
     apt-get update && \
     apt-get install -y libmsquic && \
-    rm -rf /var/lib/apt/lists/* microsoft.asc
+    rm -rf /var/lib/apt/lists/*
 
 # Build and install h2load. Required as there isn't a way to distribute h2load as a single file to download
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


docker/agent/Dockerfile
```diff
@@ -44,18 +44,19 @@ RUN apt-get update \
 RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc| sha256sum --check - && \
     apt-key add microsoft.asc && \
-    apt-add-repository https://packages.microsoft.com/debian/11/prod && \
+    rm microsoft.asc && \
+    echo deb https://packages.microsoft.com/debian/12/prod bookworm main >> /etc/apt/sources.list.d/microsoft.list && \
```
BrennanConroy, line 48: Ah yes, the ugly hack 😄

docker/agent/Dockerfile
```diff
@@ -44,18 +44,19 @@ RUN apt-get update \
...
     apt-get update && \
     apt-get install -y libmsquic && \
-    rm -rf /var/lib/apt/lists/* microsoft.asc
```
BrennanConroy, line 50: Why?

BrennanConroy (approved):

docker/agent/Dockerfile
BrennanConroy: Nvm

docker/agent/Dockerfile
```diff
@@ -44,18 +44,19 @@ RUN apt-get update \
 RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc| sha256sum --check - && \
     apt-key add microsoft.asc && \
-    apt-add-repository https://packages.microsoft.com/debian/11/prod && \
+    rm microsoft.asc && \
```
amcasey, line 47: Does it matter that this is no longer `-f`?

docker/agent/Dockerfile
amcasey: It moved above, but I'd also like to understand why.

amcasey (approved):

docker/agent/Dockerfile
sebastienros: I took these from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/debian/12/helix/amd64/Dockerfile#L41-L49

I don't know if that's necessary

docker/agent/Dockerfile
sebastienros: Maybe the other rm requires it and not this one. I tested it on 3 machines (x64 and arm64) and it worked fine.

PR #684, opened 2024-01-12
Use SetProcessAffinityMask instead of SetProcessDefaultCpuSets

sebastienros:
`JobObjectWrapper` is not necessary anymore, tested locally. We might still benefit from using a SUSPENDED thread once the API is available in the runtime.
`Environment.ProcessorCount` is also returning the correct value, and overriding its value is not necesary anymore.


EgorBo (approved):
Thanks! Looking forward for the updated crank 🙂

PR #685, opened 2024-01-12
Remove alternate spellings of "-d" and "-w" from wrk invocations.

RobertHenry6bev:
wrk can be invoked with arg "-d" or "--duration".  Likewise "-w" and "--warmup".
Allow this.
This fix isn't very elegant, and is not applied uniformly to the wrk2 and other loadgenerators.


sebastienros:
> wrk can be invoked with arg "-d" or "--duration"

You mean `wrk` or the wrk wrapper (Crank.Wrk).
Same question for `-w / --warmup`

> Allow this.

I don't understand what it's "allowing" since the previous sentence says "Can".
Can you explain what the PR is fixing or improving?

PR #689, opened 2024-01-25
Adding K6 load generator

sebastienros:


sebastienros:
This should allow us to run some benchmarks. The next step is to support custom script files to enable advanced load generation scenarios (auth, functional tests, ramp up).

src/Microsoft.Crank.Jobs.K6/Program.cs
```diff
@@ -0,0 +1,367 @@
...
+
+            args = argsList.ToArray();
+
+            string k6Url = null;
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                switch (RuntimeInformation.ProcessArchitecture)
+                {
+                    case Architecture.X64 : k6Url = "https://aspnetbenchmarks.blob.core.windows.net/tools/k6-win-amd64.exe"; break;
```
kunalspathak, line 57: no equivalent for win-arm64?

src/Microsoft.Crank.Jobs.K6/Program.cs
```diff
@@ -0,0 +1,367 @@
...
+                Console.WriteLine("A connection exception occurred while measuring the first request");
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine("An unexpected exception occurred while measuring the first request:");
+                Console.WriteLine(e.ToString());
+            }
+        }
+
+        private static string Quote(string s)
```
kunalspathak, line 345: most of the methods are in other projects as well. consider consolidating them in util project or something.

kunalspathak (approved):
LGTM. Thanks for doing this!

src/Microsoft.Crank.Jobs.K6/Program.cs
sebastienros: I would have to build it on these machines, and this is not a scenario we are looking at.

PR #690, opened 2024-01-25
Fix k6 raw summary result

sebastienros:


kunalspathak (approved):
LGTM, assuming you tested it out.

PR #696, opened 2024-02-07
Added the Startup Event to the Benchmarks Event Source

mrsharm:
Related to: https://github.com/aspnet/Benchmarks/pull/1862#issuecomment-1925110806

- Added the startup event as a part of the Benchmarks event source.
- Tested with the hello scenario using the following command: ```crank --config /crank/samples/local/local.benchmarks.yml --scenario hello --profile local --application.collect true --application.collectArguments "Providers=Benchmarks"``` 
- Observed the event showing up in the Events View in PerfView:

image

Please let me know if there is a better way to accomplish this.


sebastienros (approved):

sebastienros:
Will be deployed automatically tonight.

PR #703, opened 2024-02-24
Add measurements api

sebastienros:
When .NET can't be used it might be easier to invoke a web api instead of formatting json and writing it on the console.


MackinnonBuck (approved):
Awesome!

PR #709, opened 2024-03-07
1ES Pipelines

DrewScoggins:
This splits out the current yml AzDO definition into an internal and pr file.

The internal runs have also been configured to use the new internal templates and use the other 1ES approved steps and jobs.

The new public yml has not been tested yet, as I can't create a new pipeline from a yml file just in a branch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,17 +24,28 @@ trigger:
     - main
     - release/*
 
-pr:
-  autoCancel: true
-  branches:
-    include:
-    - '*'
+resources:
+  repositories:
+  - repository: MicroBuildTemplate
+    type: git
+    name: 1ESPipelineTemplates/MicroBuildTemplate
+    ref: refs/tags/release
+extends:
+  template: azure-pipelines/MicroBuild.1ES.Official.yml@MicroBuildTemplate
+  parameters:
+    sdl:
+      sourceAnalysisPool:
+        name: $(DncEngInternalBuildPool)
+        image: 1es-windows-2022-pt
+        os: windows
+    customBuildTags:
+    - ES365AIMigrationTooling
 
 stages:
 - stage: build
   displayName: Build
   jobs:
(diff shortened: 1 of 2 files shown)


eng/common/SetupNugetSources.ps1
```diff
@@ -35,7 +35,7 @@ Set-StrictMode -Version 2.0
 . $PSScriptRoot\tools.ps1
 
 # Add source entry to PackageSources
-function AddPackageSource($sources, $SourceName, $SourceEndPoint, $creds, $Username, $Password) {
+function AddPackageSource($sources, $SourceName, $SourceEndPoint, $creds, $Username, $pwd) {
```
sebastienros, line 38: Were/Are these changes supposed to come from arcade at some point?

sebastienros (approved):

eng/common/SetupNugetSources.ps1
DrewScoggins: They were already checked in [https://github.com/dotnet/arcade/pull/14553](here). I just copy/pasted the file from the Arcade repo to clear the SDL failures in the run.

azure-pipelines.yml
```diff
@@ -24,162 +25,106 @@ trigger:
...
-      #       pathtoPublish: artifacts/TestResults/$(_BuildConfig)/
-      #       artifactName: $(Agent.Os)_$(Agent.JobName) TestResults
-      #       artifactType: Container
-      #       parallel: true
+      - template: /eng/common/templates-official/jobs/jobs.yml@self
+        parameters:
+          enableMicrobuild: true
+          enablePublishBuildArtifacts: true
+          enablePublishTestResults: true
+          enablePublishBuildAssets: false
```
riarenas, line 54: I believe this is what's causing your BAR publishing (and release configs) to not happen in your test builds @DrewScoggins 

```suggestion
          enablePublishBuildAssets: true
```

Issue #710, opened 2024-03-08
Runs for TFM 9.0 are broken (`dotnet-install could not install a component: SDK 'edge'`)


CarnaViire:
I assume this could a regression after https://github.com/dotnet/crank/pull/708
```
> crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --scenario json --profile aspnet-perf-lin --application.framework net9.0

[03:45:44.912] Running session '616d43e81a19428a8ed3686c2284d70b' with description ''
[03:45:45.287] Starting job 'application' ...
[03:45:45.454] Submitted job: http://asp-perf-lin:5001/jobs/6
[03:45:46.977] 'application' has been selected by the server ...
[03:45:47.141] 'application' is now building ... http://asp-perf-lin:5001/jobs/6/buildlog
[03:45:49.964] 'application' failed on agent, stopping...
dotnet-install could not install a component: SDK 'edge'
[03:45:50.291] Job has failed, interrupting benchmarks ...
[03:45:50.293] Stopping job 'application' ...
[03:45:51.790] Deleting job 'application' ...

| application |   |
| ----------- | - |
```

cc @sebastienros

PR #711, opened 2024-03-08
Fix version resolution

sebastienros:
Fixes https://github.com/dotnet/crank/issues/710


CarnaViire (approved):

sebastienros:
/cc @kasperk81 

The previous PR broke "edge" versions resolution (net9.0) runs as it was not assigning the values to the result variable. This PR fixes it and also uses the same file for "latest" frameworks (aspnetcore and runtime). This way the frameworks are sure to exist and should fix random download issues.

PR #722, opened 2024-03-15
Catch unexpected exception when starting up

sebastienros:


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -923,11 +923,29 @@ private static async Task ProcessJobs(string hostname, string dockerHostname, Ca
...
+                                                {
+                                                    job.Error = "Error while starting the application: " + e.ToString();
+                                                    Log.Info(job.Error);
+
+                                                    if (job.State != JobState.Deleting &&
+                                                        job.State != JobState.Deleted &&
+                                                        job.State != JobState.Failed &&
+                                                        job.State != JobState.Stopped
+                                                    )
+                                                    {
```
mrsharm, line 944: Worth logging the exception here?

sebastienros: It's logged on line 937 (server-side) and reported to the controller on 936.

mrsharm (approved):
LGTM! Thanks

PR #726, opened 2024-03-31
Copy Files to Docker

mrsharm:
This PR implements the functionality of copying files to a docker container. Changes required to the current implementation include to create, copy the attached files and then start rather than simply running the container.

One caveat here was that the destination path should be the folder name where the files must go rather than the path to the files themselves (CC: @sebastienros). For example:

``crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/containers.benchmarks.yml --scenario json_aspnet_current --profile local --profile noload --application.options.outputFiles "./docs/development.md;app/development.md"`` 

doesn't work as it leads to the filename as: app\development.md\development.md
image

Also, quotations around the source;[destination] are necessary unlike the typical case of simply uploading a file.

NOTE: The following uploads the file to the / folder:

``crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/containers.benchmarks.yml --scenario json_aspnet_current --profile local --profile noload --application.options.outputFiles ./docs/development.md``.

## Testing

Run from the crank folder:
``crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/containers.benchmarks.yml --scenario json_aspnet_current --profile local --profile noload  --application.options.outputFiles "./docs/*;app/"``

Copied all the md files in doc folder to the container:
image

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -2164,13 +2164,11 @@ namespace Microsoft.Crank.Agent
 
             var containerName = Regex.Replace(imageName, @"[^\w]", "_")+ $"-{job.Id}";
 
-            // TODO: Clean previous images 
-
             // Stop container in case it failed to stop earlier
-            // await ProcessUtil.RunAsync("docker", $"stop {containerName}", throwOnError: false);
+            await ProcessUtil.RunAsync("docker", $"stop {containerName}", throwOnError: false);
 
             // Delete container if the same name already exists
-            // await ProcessUtil.RunAsync("docker", $"rm {imageName}", throwOnError: false);
+            await ProcessUtil.RunAsync("docker", $"rm {imageName}", throwOnError: false);
 
             if (!String.IsNullOrWhiteSpace(job.CpuSet))
             {
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -2192,7 +2192,6 @@ namespace Microsoft.Crank.Agent
 
             var createCommandResult = await ProcessUtil.RunAsync("docker", $"{createCommand} ",
                 throwOnError: true,
-                onStart: _ => stopwatch.Start(),
                 captureOutput: true,
                 log: true,
                 outputDataReceived: job.BuildLog.AddLine
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -2187,16 +2187,62 @@ namespace Microsoft.Crank.Agent
                 environmentArguments += $"--memory=\"{job.MemoryLimitInBytes}b\" ";
             }
 
-            var command = $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --privileged --network host {imageName} {job.DockerCommand}";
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2187,16 +2187,47 @@ private static async Task<(string containerId, string imageName, string workingD
...
+            foreach (var attachment in job.Attachments)
+            {
+                var filename = attachment.Filename.Replace("\\", "/");
+                var tempFilePath = attachment.TempFilename;
+
+                Log.Info($"Copying output file to container: {filename}");
+
+                string dockerCopyCommand = $"cp {tempFilePath} {containerName}:{filename}";
+                var copyResult = await ProcessUtil.RunAsync("docker", $"{dockerCopyCommand} ",
+                    throwOnError: true,
```
cshung, line 2213: Should we worry about stale container created if this fails?

mrsharm: Yes, definitely - guarding against that now.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2187,16 +2185,47 @@ private static async Task<(string containerId, string imageName, string workingD
...
+                    onStart: _ => stopwatch.Start(),
```
sebastienros, line 2212:
```suggestion
```
Nothing to track here

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2187,16 +2185,47 @@ private static async Task<(string containerId, string imageName, string workingD
...
 
-            var command = $"run -d {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --privileged --network host {imageName} {job.DockerCommand}";
+            // docker create --name {containerName}
+            var createCommand = $"create {environmentArguments} {job.Arguments} --label benchmarks --name {containerName} --privileged --network host {imageName} {job.DockerCommand}";
+
+            job.BuildLog.AddLine("docker " + createCommand);
+
+            var createCommandResult = await ProcessUtil.RunAsync("docker", $"{createCommand} ",
+                throwOnError: true,
+                onStart: _ => stopwatch.Start(),
```
sebastienros, line 2195:
```suggestion
```
Nothing to track here

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2187,16 +2185,47 @@ private static async Task<(string containerId, string imageName, string workingD
...
             }
 
-            job.BuildLog.AddLine("docker " + command);
+            var startCommand = $"start {containerName}";
+            job.BuildLog.AddLine("docker " + startCommand);
 
-            var result = await ProcessUtil.RunAsync("docker", $"{command} ",
+            var result = await ProcessUtil.RunAsync("docker", $"{startCommand} ",
                 throwOnError: true,
                 onStart: _ => stopwatch.Start(),
```
sebastienros, line 2230: This 'start' is fine though because we'll track the startup time from here

sebastienros:
We'll need a comment in the docs to explain that in docker the destination path is the final destination in the container.

sebastienros (approved):

sebastienros:
@mrsharm do you want to merge this PR or are you waiting to add some more things?

mrsharm:
Thanks - sorry for the delay. Merged now.

PR #729, opened 2024-04-22
Fix touch endpoint to not update last communication datetime

sebastienros:


src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -52,23 +51,6 @@ bool IsActive(Job job)
             }
         }
 
-        [HttpGet("legacy/{id}/touch")]
```
sebastienros, line 55: Removing legacy endpoints too. It's using `JobApis.cs` for a long time now.

caaavik-msft (approved):

PR #730, opened 2024-04-22
Convert to central package management

sebastienros:


NuGet.config
```diff
@@ -16,5 +16,15 @@
     <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
     <add key="dotnet5-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5-transport/nuget/v3/index.json" />
   </packageSources>
-  <disabledPackageSources />
+  <packageSourceMapping>
+    <packageSource key="dotnet-public">
+      <package pattern="*" />
```
joperezr, line 21: NIT: If you are only requiring these three package source mappings, it means that all of the feeds that you have under  are not needed and can be cleaned up.

Directory.Packages.props
```diff
@@ -0,0 +1,46 @@
...
+    <PackageVersion Include="Parlot" Version="0.0.25" />
+    <PackageVersion Include="Serilog.AspNetCore" Version="4.1.0" />
+    <PackageVersion Include="Serilog.Formatting.Compact" Version="1.1.0" />
+    <PackageVersion Include="Serilog.Sinks.Console" Version="4.0.0" />
+    <PackageVersion Include="Serilog.Sinks.File" Version="5.0.0" />
+    <PackageVersion Include="System.CommandLine" Version="2.0.0-beta1.20427.1" />
+    <PackageVersion Include="System.IO.Hashing" Version="8.0.0" />
+    <PackageVersion Include="System.IO.Pipelines" Version="4.6.0" />
+    <PackageVersion Include="System.Linq.Async" Version="4.1.1" />
+    <PackageVersion Include="System.Text.Encodings.Web" Version="4.7.2" AllowExplicitReference="true" />
```
joperezr, line 42: Interesting, first time I see the AllowExplicitReference. Do you know what is that for?

src/Microsoft.Crank.Models/Microsoft.Crank.Models.csproj
```diff
@@ -2,12 +2,12 @@
...
   <PropertyGroup>
     <Description>Class to transfer data to/from the benchmark server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
-    <PackageReference Include="Microsoft.AspNetCore.Http.Features" Version="3.1.8" />
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
joperezr, line 9: Should we instead change the SDK to use the aspnet one so it gets this ref for free?

src/Microsoft.Crank.Models/Microsoft.Crank.Models.csproj
```diff
@@ -2,12 +2,12 @@
 
   <PropertyGroup>
     <Description>Class to transfer data to/from the benchmark server.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
joperezr, line 5: Is this a project that gets used anywhere else outside the repo? Mainly asking as this is a change that is obviously breaking.

joperezr (approved):
Left minor comments, but this looks great otherwise.

src/Microsoft.Crank.Models/Microsoft.Crank.Models.csproj
sebastienros: Only in this repos, not published on NuGet.

src/Microsoft.Crank.Models/Microsoft.Crank.Models.csproj
sebastienros: It asks for a Main method at that point.

joperezr: I think you should be able to set `OutputType` to `Library` and that should fix this, but anyway, just a NIT, so feel free to disregard.

Directory.Packages.props
sebastienros: https://github.com/dotnet/aspnetcore/blob/main/eng/targets/ResolveReferences.targets#L68

Talking with Brennan it could be related to the fact that it's part of the shared framework, and that it could complain otherwise?

PR #734, opened 2024-05-01
Use platform specific product commit url

sebastienros:


sebastienros:
Files from https://github.com/dotnet/installer?tab=readme-ov-file#table
Examples:
https://aka.ms/dotnet/9.0.1xx/daily/productCommit-linux-x64.txt
https://aka.ms/dotnet/9.0.1xx/daily/productCommit-win-x64.txt

These files used to have the same SDK version, since the last build they don't

LoopedBard3 (approved):
Looks good to me.

PR #735, opened 2024-05-09
Add certificate based authentication

DrewScoggins:
This PR add support for using certificates to authenticate for the Crank controller, agent, and ServiceBusAgent. It does not add support for using certificates to authenticate Azure relay.


sebastienros:
@DrewScoggins  can you check my feedback?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -349,6 +382,35 @@ private static async Task<int> Run(string url, string hostname, string dockerHos
...
+                    ClientCertificateCredential ccc = null;
+                    X509Store store = null;
+                    X509Certificate2 certificate = null;
+                    if (_certBasedAuth.HasValue())
+                    {
+                        for (int i = 1; i <= 8; i++)
+                        {
+                            store = new X509Store((StoreName)i, StoreLocation.LocalMachine);
+                            store.Open(OpenFlags.ReadOnly);
+                            foreach (var cert in store.Certificates)
```
caaavik-msft, line 394: This for loop (and the others that do the same thing) can be replaced with a call to `store.Certificates.Find(X509FindType.FindByThumbprint, _certThumbprint.Value(), false)`

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -48,15 +57,64 @@ public static int Main(string[] args)
                     queue = Environment.GetEnvironmentVariable(queue);
                 }
 
-                await ProcessAzureQueue(connectionString, queue);
+                if (certBasedAuth.HasValue() && !(certClientId.HasValue() && certTenantId.HasValue() && certThumbprint.HasValue()))
```
caaavik-msft, line 60: This doesn't handle the case where someone passes a cert path instead of a thumbprint (not sure if there is a need for both to be configured at once). Is the goal of having certPath as an option to make it so that we can rotate certificates without having to modify the args passed to crank?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -48,15 +57,64 @@ public static int Main(string[] args)
...
+                {
+                    await ProcessAzureQueue(connectionString, queue);
+                }
             });
 
             return app.Execute(args);
         }
 
-        private static async Task ProcessAzureQueue(string connectionString, string queue)
+        private static async Task ProcessAzureQueue(string connectionString, string queue, bool certBasedAuth = false, string certPath = "", string certThumbprint = "", string certTenantId = "", string certClientId = "")
```
caaavik-msft, line 79: Maybe we should create a reusable type that puts all these certificate related options into a single class and then we just need to pass around that object instead, should make things a little cleaner. This would eliminate the need for a boolean to indicate if certBasedAuth should be used since we can just use `null` to represent that.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -358,11 +420,15 @@ private static async Task<int> Run(string url, string hostname, string dockerHos
 
                     var rcsb = new RelayConnectionStringBuilder(relayConnectionString);
 
+                    if (_certBasedAuth.HasValue())
+                    {
+                        options.TokenProvider = GetAadTokenProvider(_certTenantId.Value(), _certClientId.Value(), certificate);
```
caaavik-msft, line 425: GetAadTokenProvider takes clientId, then tenantId as parameter order, so it is passed in incorrect order here

src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -3351,7 +3388,11 @@ private static async Task<string> GetRelayTokenAsync(Uri endpointUri)
             }
 
             var rcsb = new RelayConnectionStringBuilder(connectionString);
-            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(rcsb.SharedAccessKeyName, rcsb.SharedAccessKey);
+            TokenProvider tokenProvider = null;
+            AccessToken token = default;
+            
+            tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(rcsb.SharedAccessKeyName, rcsb.SharedAccessKey);
```
caaavik-msft, line 3394: This should make use of GetAadTokenBuilder, and I think we should do the same for the service bus, perhaps we can make the TokenProvider code reusable across the projects?

caaavik-msft (changes requested):

src/Microsoft.Crank.Controller/Program.cs
sebastienros: What is GetAadTokenBuilder, did you mean GetAadTokenProvider?

caaavik-msft: ah yes, I meant `GetAadTokenProvider `

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -349,6 +382,35 @@ private static async Task<int> Run(string url, string hostname, string dockerHos
...
+                    {
+                        for (int i = 1; i <= 8; i++)
+                        {
+                            store = new X509Store((StoreName)i, StoreLocation.LocalMachine);
+                            store.Open(OpenFlags.ReadOnly);
+                            foreach (var cert in store.Certificates)
+                            {
+                                if (cert.Thumbprint == _certThumbprint.Value())
+                                {
+                                    ccc = new ClientCertificateCredential(_certTenantId.Value(), _certClientId.Value(), cert);
```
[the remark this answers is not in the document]
DrewScoggins: Yeah, this was leftover code from when I thought I had to have a credential here.

src/Microsoft.Crank.Controller/JobSerializer.cs
```diff
@@ -11,18 +11,24 @@
...
         public static Task WriteJobResultsToSqlAsync(
             JobResults jobResults, 
-            string sqlConnectionString, 
+            string sqlConnectionString,
             string tableName,
             string session,
             string scenario,
-            string description
+            string description,
+            bool certBasedAuth = false,
```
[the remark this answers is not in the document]
DrewScoggins: Yes, these parameters were not finished. I have added in the rest and passed them along to the `WriteResultsToSql` method

src/Microsoft.Crank.Controller/JobSerializer.cs
```diff
@@ -216,6 +234,51 @@ JobResults jobResults
...
+                        }
+                    }
+                    if (ccc != null)
+                    {
+                        break;
+                    }
+                }
+                if(ccc == null)
+                {
+                    ccc = new ClientCertificateCredential(certTenantId, certClientId, certPath);
```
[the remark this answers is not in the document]
DrewScoggins: Makes sense.

src/Microsoft.Crank.Controller/JobSerializer.cs
```diff
@@ -83,7 +96,12 @@ static async Task InitializeDatabaseInternalAsync(string connectionString, strin
             string session,
             string scenario,
             string description,
-            string document
+            string document,
+            bool certBasedAuth = false,
```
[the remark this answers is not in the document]
DrewScoggins: By "load the cert" do you mean create the `ClientCertificateCredential` at the beginning and pass that around? Even if we load the cert, we still would need to at least pass around the tenant and client id. Admittedly, three things instead of five being passed around is fewer, but it won't be as simple as passing around the cert.

src/Microsoft.Crank.Controller/Program.cs
```diff
@@ -3351,7 +3388,11 @@ private static async Task<string> GetRelayTokenAsync(Uri endpointUri)
             }
 
             var rcsb = new RelayConnectionStringBuilder(connectionString);
-            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(rcsb.SharedAccessKeyName, rcsb.SharedAccessKey);
+            TokenProvider tokenProvider = null;
+            AccessToken token = default;
```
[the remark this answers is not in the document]
DrewScoggins: I forgot to remove this one I removed the cert-based auth support on the controller side for Azure Relay. I did this because conditional access made it where you could only use cert-based auth on computers that could already access the resources that you cared about.

src/Microsoft.Crank.Agent/Startup.cs
DrewScoggins: Is there a reason that we would want to return invalid certificates?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
DrewScoggins: I added it because I was wanting to make sure that we were giving maximum flexibility in how people could use the tool. Good catch on missing that case. I have updated the expression so that it should do what we want.

src/Microsoft.Crank.Agent/Startup.cs
DrewScoggins: I changed the definition of `GetAadTokenProvider` so that it takes the tenantId first. This matches how all of the other APIs are written.

src/Microsoft.Crank.Controller/Program.cs
DrewScoggins: So as I mentioned in one of Sebastien's comments, I removed the use of certs for the Controller in Azure Relay. This is because, due to conditional access policy, we can only use the service principal authentication story on machines that can access the machines on the relay. So for now I was thinking that we would not change this behavior.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
DrewScoggins: This likely makes the most sense.

DrewScoggins:
Sorry for the delay in responding to this feedback, but I think I have addressed all the feedback. This PR is also not complete because we need to also add managed identity support, as that is how the ASP.NET lab is going to function.

mkArtakMSFT:
> Sorry for the delay in responding to this feedback, but I think I have addressed all the feedback. This PR is also not complete because we need to also add managed identity support, as that is how the ASP.NET lab is going to function.

Thanks for the update, @DrewScoggins. @sebastienros do we need a change from our side here, or should this merge first and then we follow-up with a separate PR?

MackinnonBuck (approved):

Issue #756, opened 2024-07-29
ResolveXXXVersion should not use CurrentXXXVersion


sebastienros:
Each method should download the release-metadata file if necessary and take appropriate action if the file doesn't exist (for new frameworks that haven't had a release yet). In that case otherwise the `currentXXXVersion` argument is null and causes errors.

The file could be cached locally to prevent from downloading it multiple times per benchmark and instances.

Related to https://github.com/dotnet/crank/pull/754

PR #757, opened 2024-07-29
Refactor CurrentVersions and cache file downloads

sebastienros:
Fixes https://github.com/dotnet/crank/issues/756

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -35,6 +35,7 @@ using Microsoft.Crank.Models.Security;
 using Microsoft.Diagnostics.NETCore.Client;
 using Microsoft.Diagnostics.Tools.Trace;
 using Microsoft.Diagnostics.Tracing;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Identity.Client;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -132,8 +135,9 @@ Based on the target framework
         private static Process perfCollectProcess;
         private static readonly object _synLock = new();
         private static object _consoleLock = new();
+        private static MemoryCache _fileContentCache = new(new MemoryCacheOptions { SizeLimit = 10_000_000 });
 
-        private static Task dotnetTraceTask;
+private static Task dotnetTraceTask;
```
halter73, line 140:
```suggestion
        private static Task dotnetTraceTask;
```

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3835,24 +3837,28 @@ private static string SeekCompatibleDesktopRuntime(string dotnetHome, string tar
...
+
                     Log.Info($"ASP.NET: {aspNetCoreVersion} (Current)");
                     break;
                 case "latest":
                     // aspnet runtime service releases are not published on feeds
                     switch (versionPrefix)
                     {
                         case "9.0":
-                            var productsInfo = JObject.Parse(await DownloadContentAsync(_latestProductVersions90Url));
+                            var productsInfo = JObject.Parse(await DownloadContentAsync(_latestProductVersions90Url, cache: _latestProductVersions90Cache));
```
halter73, line 3861: It looks like `_latestProductVersions90Cache` is the actual memory cache here. I would rename both the static and the parameter to include timeout.

```suggestion
                            var productsInfo = JObject.Parse(await DownloadContentAsync(_latestProductVersions90Url, cacheTimeout: _latestProductVersions90CacheTimeout));
```

Same goes for `_aspnetSdkVersionCache` -> `_aspnetSdkVersionCacheTimeout`.

halter73 (approved):

joperezr:
FYI @sebastienros PR was merged due to automerge but @halter73 had some feedback above. Commenting to make sure that isn't missed 😉

sebastienros:
I saw that, thanks

PR #758, opened 2024-08-07
Shut down MSBuild server after builds

sebastienros:
VB/C# Compiler runs with ServerGC and impacts some benchmark latencies.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -3261,7 +3261,7 @@ namespace Microsoft.Crank.Agent
 
                 var projectName = Path.GetFileName(FormatPathSeparators(job.Project));
 
-                var arguments = $"publish {projectName} -c Release -o {outputFolder} {buildParameters}";
+                var arguments = $"publish {projectName} --disable-build-servers -c Release -o {outputFolder} {buildParameters}";
 
                 // This might be set already, and the SDK will then use it for some targets files
                 // https://github.com/dotnet/sdk/blob/e2faebad758a7d38b5965cda755a17e9e9881599/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs#L75
@@ -3311,15 +3311,6 @@ namespace Microsoft.Crank.Agent
                 Log.Info($"Application published successfully in {job.BuildTime.TotalMilliseconds} ms");
 
                 PatchRuntimeConfig(job, outputFolder, aspNetCoreVersion, runtimeVersion);
-
-                await ProcessUtil.RunAsync(dotnetExecutable, "build-server shutdown",
-                    workingDirectory: benchmarkedApp,
-                    environmentVariables: env,
-                    throwOnError: false,
-                    outputDataReceived: job.BuildLog.AddLine,
-                    cancellationToken: cancellationToken,
-                   timeout: TimeSpan.FromSeconds(20)
-                );
             }
 
             var publishedSize = DirSize(new DirectoryInfo(outputFolder)) / 1024;
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -3319,7 +3319,7 @@ namespace Microsoft.Crank.Agent
                     outputDataReceived: job.BuildLog.AddLine,
                     cancellationToken: cancellationToken,
                    timeout: TimeSpan.FromSeconds(20)
-                    );
+                );
             }
 
             var publishedSize = DirSize(new DirectoryInfo(outputFolder)) / 1024;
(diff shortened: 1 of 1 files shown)


caaavik-msft:
If there is another process that is building at the same time, then would this cause it to kill that build? One other option aside from shutting down the build server would be to run the builds with `MSBUILDDISABLENODEREUSE=1` that way it can still make use of shared compilation but won't keep the nodes up longer after the compilation is completed. This is the solution we use for the dotnet/performance runs.

mrsharm:
> If there is another process that is building at the same time, then would this cause it to kill that build? One other option aside from shutting down the build server would be to run the builds with `MSBUILDDISABLENODEREUSE=1` that way it can still make use of shared compilation but won't keep the nodes up longer after the compilation is completed. This is the solution we use for the dotnet/performance runs.

Curious, why would we want to permit building _while_ running a benchmark?

caaavik-msft:
For scenarios where the crank agent is running the benchmark in production where the goal is to measure CPU usage, then yes it wouldn't make sense. But sometimes the agent is running something that doesn't need CPU measured (e.g. load generator for load test), or it is running on a local machine during development where we don't care about accurate results and just want to test that the job is working, and so this would shutdown the build server on my local machine.

sebastienros:
> But sometimes the agent is running something that doesn't need CPU measured

Question is "does it hurt to do it" and in that case it's ok

> it is running on a local machine during development

Could be a problem, but not sure about the actual impact even in that case.

Nonetheless, I will check it there is a way to do `dotnet run` without starting the build server.

Note that this came from feedback from the GC team as they were seeing some peak latencies due to ServerGC while the app they measured didn't have it running. It happens that this VBCSCompiler process was triggering collections during an unrelated benchmark.

sebastienros:
Testing `--disable-build-servers` now

mrsharm:
Thanks for working on this, @sebastienros! I am observing that we no longer observe any induced GCs (which was the original issue) for the dotnet build process for all cases except FortunesEf based on a recent run.

Is there something different about FortunesEf in terms of building?

## Details

That dotnet process seems to be with the following commandline:

"C:\Users\Administrator\AppData\Local\Temp\benchmarks-agent\benchmarks-server-6984\jsaa1fxx.i2y\dotnet.exe" exec "C:\Users\Administrator\AppData\Local\Temp\benchmarks-agent\benchmarks-server-6984\jsaa1fxx.i2y\sdk\9.0.100-rc.1.24407.33\Roslyn\bincore\VBCSCompiler.dll"

The parent process didn't seem to a part of the trace. 

### Details of the Induced GCs for FortunesEf (All Gen2 Blocking are still Induced)

```
Q:\DATAS_Fixes\PerfCheck\datas_28c_baseline_0\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 1
Q:\DATAS_Fixes\PerfCheck\datas_28c_baseline_1\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 3
Q:\DATAS_Fixes\PerfCheck\datas_28c_baseline_2\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 2
Q:\DATAS_Fixes\PerfCheck\datas_28c_baseline_3\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 1
Q:\DATAS_Fixes\PerfCheck\datas_28c_fix_0\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 2
Q:\DATAS_Fixes\PerfCheck\datas_28c_fix_1\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 1
Q:\DATAS_Fixes\PerfCheck\datas_28c_fix_2\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 1
Q:\DATAS_Fixes\PerfCheck\svr_28c_0\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 3
Q:\DATAS_Fixes\PerfCheck\svr_28c_1\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 3
Q:\DATAS_Fixes\PerfCheck\svr_28c_2\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 2
Q:\DATAS_Fixes\PerfCheck\svr_28c_3\FortunesEf_Windows.gc.etl.zip: dotnet 7700: # of Gen2 Blocking GCs: 2
```

This is affecting all iterations i.e., it's pervasive across all runs.

sebastienros:
What could explain that is that FortunesEF has some Razor views, which are compiled by ASP.NET and maybe the way it's done doesn't preserve the `--disable-build-servers` flag.

Let's try with the ENV as initially suggested. This will get deployed tonight automatically on citrine-win.

mrsharm (approved):
Just tested this change and no longer observe any induced GCs on citrine. Thanks!

PR #762, opened 2024-09-20
Map cgroupfs with Docker to fix cgroups v2 support

sebastienros:


joperezr (approved):
LGTM, thanks @sebastienros

PR #774, opened 2024-11-26
Add support for Ultra

sebastienros:


DeagleGross (approved):

DeagleGross:
We should add docs for Ultra usage as well (`--application.profile true --application.profileType ultra`)

sebastienros:
Will add the docs. Also the tests are failing which is not right, they are quite stable. The tool also just got support for output filenames, so this can be used instead of taking whatever 'json.gz' is in the folder.

DeagleGross (approved):

PR #777, opened 2024-12-02
Detect Azure Relay certificate authentication in docker script

sebastienros:


LoopedBard3 (approved):
Looks good. Glad we are getting through this 👍.

DrewScoggins:
LGTM. And as Sebastien mentioned since these machines are now running from inside the trusted boundary we don't have to worry about conditional access policy.

DrewScoggins (approved):

PR #786, opened 2025-01-20
feat: support SslProtocol parameter for httpClient job

DeagleGross:
Adding an option to explicitly set the sslProtocol version to use on `SocketsHttpHandler`

### DIFF
--- a/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
+++ b/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
@@ -596,7 +596,7 @@ namespace Microsoft.Crank.Jobs.HttpClientClient
             var maxLatency = 0D;
             var totalLatency = 0D;
             var transferred = 0L;
-            var measuringStart = 0L;
+            var measuringStart = -1L;
 
             var sw = new Stopwatch();
             sw.Start();
@@ -652,7 +652,7 @@ namespace Microsoft.Crank.Jobs.HttpClientClient
                     {
                         if (measuringStart == 0)
                         {
-                            measuringStart = sw.ElapsedTicks;
+                            measuringStart = sw.ElapsedTicks - 1;
                         }
 
                         transferred += responseMessage.Content.Headers.ContentLength ?? 0;
@@ -709,16 +709,7 @@ namespace Microsoft.Crank.Jobs.HttpClientClient
                 }
             }
 
-            long throughput = 0;
-            try
-            {
-                throughput = transferred / ((sw.ElapsedTicks - measuringStart) / Stopwatch.Frequency);
-            }
-            catch (DivideByZeroException ex)
-            {
-                Log("Error during throughput calculation: {0}", ex.Message);
-            }
-
+            var throughput = transferred / ((sw.ElapsedTicks - measuringStart) / Stopwatch.Frequency);
 
             if (!String.IsNullOrWhiteSpace(Script) && !worker.Script.GetValue("stop").IsUndefined())
             {
--- a/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
+++ b/src/Microsoft.Crank.Jobs.HttpClient/Program.cs
@@ -184,27 +184,28 @@ namespace Microsoft.Crank.Jobs.HttpClientClient
 
                 Headers = new List<string>(optionHeaders.Values);
 
-                var sslProtocols = ParseSslProtocols(optionSslProtocols.Values);
-                _httpClientHandler.SslProtocols = sslProtocols;
-                Log("Using sslProtocols: {0}. input: {1}", sslProtocols, string.Join(";", optionSslProtocols.Values));
-
-                static SslProtocols ParseSslProtocols(List<string> inputProtocols)
+                if (optionSslProtocols.HasValue())
                 {
-                    if (inputProtocols is null)
+                    var sslProtocols = ParseSslProtocols(optionSslProtocols.Values);
+                    if (sslProtocols is not SslProtocols.None)
                     {
-                        return SslProtocols.Tls12 | SslProtocols.Tls13;
+                        Log("Using sslProtocols: {0}", sslProtocols);
+                        _httpClientHandler.SslProtocols = sslProtocols;
                     }
 
-                    var result = SslProtocols.None;
-                    foreach (var protocol in inputProtocols)
+                    static SslProtocols ParseSslProtocols(List<string> inputProtocols)
                     {
-                        if (Enum.TryParse<SslProtocols>(protocol, ignoreCase: true, out var mask))
+                        var result = SslProtocols.None;
+                        foreach (var protocol in inputProtocols)
                         {
-                            result |= mask;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -683,7 +716,16 @@ public static async Task<WorkerResult> DoWorkAsync()
                 }
             }
 
-            var throughput = transferred / ((sw.ElapsedTicks - measuringStart) / Stopwatch.Frequency);
+            long throughput = 0;
+            try
+            {
+                throughput = transferred / ((sw.ElapsedTicks - measuringStart) / Stopwatch.Frequency);
```
DeagleGross, line 722: I was running the following job (needs adjustment for running without local changes)
> crank --config https://raw.githubusercontent.com/DeagleGross/Benchmarks/refs/heads/dmkorolev/idna/tls-protocols/scenarios/tls.benchmarks.yml --scenario tls-handshakes-httpsys --profile aspnet-gold-win --application.framework net10.0 --application.options.outputFiles "D:\code\Benchmarks\src\BenchmarksApps\TLS\HttpSys\bin\Release\net9.0\*.dll" --application.variables.logRequestDetails "true" --load.variables.sslProtocols "tls13"

and it fails with such exception (below), therefore I decided to handle it. Not sure why that happens only for tls13 scenario.

```
[STDERR] Unhandled exception. System.DivideByZeroException: Attempted to divide by zero.
[STDERR]    at Microsoft.Crank.Jobs.HttpClientClient.Program.DoWorkAsync() in /tmp/benchmarks-agent/benchmarks-server-1/pzqitig2.gjs/src/Microsoft.Crank.Jobs.HttpClient/Program.cs:line 718
[STDERR]    at Microsoft.Crank.Jobs.HttpClientClient.Program.RunAsync() in /tmp/benchmarks-agent/benchmarks-server-1/pzqitig2.gjs/src/Microsoft.Crank.Jobs.HttpClient/Program.cs:line 388
[STDERR]    at Microsoft.Crank.Jobs.HttpClientClient.Program.<>c__DisplayClass86_0.<<Main>b__1>d.MoveNext() in /tmp/benchmarks-agent/benchmarks-server-1/pzqitig2.gjs/src/Microsoft.Crank.Jobs.HttpClient/Program.cs:line 295
[STDERR] --- End of stack trace from previous location ---
[STDERR]    at McMaster.Extensions.CommandLineUtils.CommandLineApplicationExtensions.<>c__DisplayClass8_0.<<OnExecuteAsync>b__0>d.MoveNext()
[STDERR] --- End of stack trace from previous location ---
[STDERR]    at McMaster.Extensions.CommandLineUtils.CommandLineApplication.ExecuteAsync(String[] args, CancellationToken cancellationToken)
[STDERR]    at Microsoft.Crank.Jobs.HttpClientClient.Program.Main(String[] args) in /tmp/benchmarks-agent/benchmarks-server-1/pzqitig2.gjs/src/Microsoft.Crank.Jobs.HttpClient/Program.cs:line 298
[STDERR]    at Microsoft.Crank.Jobs.HttpClientClient.Program.<Main>(String[] args)
```

src/Microsoft.Crank.Jobs.HttpClient/httpclient.yml
```diff
@@ -28,4 +28,5 @@ jobs:
       httpVersion: 1.1
       certPath: # the path or url of the certificate used for authentication
       certPwd: # the password of the certificate specified in certPath
-    arguments: '--url "{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}" --connections {{connections}} --warmup {{warmup}} --duration {{duration}} {{headers[presetHeaders]}} --version {{httpVersion}} {% if certPath %} --cert {{certPath}} {% endif %} {% if certPwd %} --certpwd {{certPwd}} {% endif %}'
+      sslProtocols: # the SSL protocols to use, e.g. Tls12, Tls13.
+    arguments: '--url "{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}" --connections {{connections}} --warmup {{warmup}} --duration {{duration}} {{headers[presetHeaders]}} --version {{httpVersion}} {% if certPath %} --cert {{certPath}} {% endif %} {% if certPwd %} --certpwd {{certPwd}} {% endif %} {% if sslProtocols %} --sslProtocols {{sslProtocols}} {% endif %}'
```
sebastienros, line 32:
```suggestion
      sslProtocol: # the SSL protocol to use, e.g. Tls12, Tls13.
    arguments: '--url "{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}" --connections {{connections}} --warmup {{warmup}} --duration {{duration}} {{headers[presetHeaders]}} --version {{httpVersion}} {% if certPath %} --cert {{certPath}} {% endif %} {% if certPwd %} --certpwd {{certPwd}} {% endif %} {% if sslProtocol %} --ssl-protocol {{sslProtocol}} {% endif %}'
```

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
BrennanConroy: `DivideByZeroException` is one of the exception types you generally shouldn't be explicitly catching. We should instead figure out why it happened and fix that, or change the logic so we avoid dividing by zero in the first place.

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
```diff
@@ -182,6 +185,29 @@ static async Task Main(string[] args)
...
+                        foreach (var protocol in inputProtocols)
+                        {
+                            if (Enum.TryParse<SslProtocols>(protocol, ignoreCase: true, out var mask))
+                            {
+                                result |= mask;
+                            }
+                        }
+
+                        return result;
+                    }
```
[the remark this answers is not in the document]
DeagleGross: moved under `app.OnValidate(...)` to be able to return `ValidationResult`

src/Microsoft.Crank.Jobs.HttpClient/Program.cs
sebastienros: `sw.ElapsedTicks` could be equal to `measuringStart` if:
- they are both not changes (default is 0)
- code is so fast or optimized that `sw.ElapsedTicks` is not updated after `measuringStart` is assigned

To solve this we could initialize `measuringStart` with `-1` and also set `measuringStart = sw.ElapsedTicks - 1` to ensure they are never the same. The computation would be off by 1 tick, which wouldn't matter. And if they were to be identical then we can't get anything from the result so the newly non-zero value is as meaningless. While no exception will be thrown.

DeagleGross: makes sense, I adjusted the code as proposed

DeagleGross: added the `(double)` cast to be 100% sure there is no division by zero

PR #787, opened 2025-01-21
preventing from injecting malicious code by JobsController/download

e-kharion:
preventing from injecting malicious code by JobsController/download
preventing from injecting malicious code by JobSerializer - using env. variable table name

### DIFF
--- a/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
+++ b/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
@@ -712,18 +712,12 @@ namespace Microsoft.Crank.Agent.Controllers
 
                     try
                     {
-                        if (!System.IO.File.Exists(fullPath))
-                        {
-                            Log.Error($"Client was requested to copy image from not existing path '{path}'");
-                            return NotFound();
-                        }
-
                         // docker cp mycontainer:/foo.txt foo.txt 
 
                         var destinationFilename = Path.Combine(tempDirectory, Path.GetFileName(path));
 
                         // Delete container if the same name already exists
-                        var result = await ProcessUtil.RunAsync("docker", $"cp {job.GetNormalizedImageName()}:{fullPath} {destinationFilename}", throwOnError: false, log: true);
+                        var result = await ProcessUtil.RunAsync("docker", [ "cp", $"{job.GetNormalizedImageName()}:{path}", destinationFilename ], throwOnError: false, log: true);
 
                         return new GZipFileResult(destinationFilename);
                     }
--- a/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
+++ b/src/Microsoft.Crank.Agent/Controllers/JobsController.cs
@@ -712,12 +712,18 @@ namespace Microsoft.Crank.Agent.Controllers
 
                     try
                     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Crank.Controller/JobSerializer.cs
```diff
@@ -99,6 +99,7 @@ CertificateOptions certificateOptions
 
             var insertCmd =
                 @"
+                IF EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND TABLE_NAME = @TableName)
```
sebastienros, line 102: What's the goal since it already creates the table if it doesn't exist? Did you find a way to make it crash on save as the table wasn't created?

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
```diff
@@ -712,12 +712,18 @@ public async Task<IActionResult> Download(int id, string path)
 
                     try
                     {
+                        if (!System.IO.File.Exists(fullPath))
```
sebastienros, line 715: It is checking against the local filesystem but `path` is used relative to the running docker container. I don't think this will ever work.

src/Microsoft.Crank.Controller/JobSerializer.cs
e-kharion: Hi @sebastienros
The concern here is purely security - seems vulnerable to SQL injection if the attacker would update the related table name env. variable, the insert command could have execute a malicious sql code. 
Updated code should prevent it as the table name is passed as parameter and the table name now cannot contain the ']' character

src/Microsoft.Crank.Agent/Controllers/JobsController.cs
e-kharion: good catch, thanks for that. I at least updated the `ProcessUtil.RunCommand` method so it internally escapes the process start arguments

src/Microsoft.Crank.Controller/JobSerializer.cs
sebastienros: The code is run with the current user's connection string and credentials. If the user want to run some malicious code then can just do it. It's not the same as if this code was run by a server, on behalf of a user.

src/Microsoft.Crank.Controller/JobSerializer.cs
```diff
@@ -55,7 +55,7 @@ public static async Task InitializeDatabaseAsync(
         {
             var createCmd =
                 @"
-                IF OBJECT_ID(N'dbo." + tableName + @"', N'U') IS NULL
+                IF NOT EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND TABLE_NAME = @TableName)
                 BEGIN
                     CREATE TABLE [dbo].[" + tableName + @"](
```
sebastienros, line 60: This one is not parameterized, so the issue is still there. But this can't be parameterized, so you can updated your changes with something like this:

```c#
var encodedTableName = new SqlCommandBuilder().QuoteIdentifier(unquotedIdentifier: tableName);
```

sebastienros (approved):

PR #790, opened 2025-01-31
feat: collect and report machine-level stats (cpu usage)

DeagleGross:
Implementing a request to collect machine-level stats (at least cpu usage) for windows and linux:
- Windows implementation: collect [% Processor Time / _Total](https://learn.microsoft.com/en-us/windows/win32/perfctrs/collecting-performance-data) with intervals of 1 sec
- Linux implementation: collect [vmstat](https://www.redhat.com/en/blog/linux-commands-vmstat) periodically and parse the output.

Additional features:
1) crank has support for `--cpuSet`, which restricts the processors used for running the child process. For linux it is not supported, but for windows I am spawning metric collection for the specified processors, and there is an additional result `Machine Total Cpu-Set Usage (%)` reported.

2) if new flag `job.options.collectMachineCounters` is specified, some known processes metrics will also be collected (for now only `lsass` on windows is supported).

tested on both wsl (linux) and windows. Below is the result of run with `--job.options.collectMachineCounters` and `--job.cpuSet "0"` arguments. It shows that process-cpu is on the level (>90%) with cpu-set usage (because a single processor is used to run a child process), but total usage is only 11% because of my local multi-processor machine
image

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,7 +28,7 @@
     <PackageVersion Include="Microsoft.Windows.CsWin32" Version="0.3.49-beta" PrivateAssets="all" />
     <PackageVersion Include="Mono.Cecil" Version="0.11.3" />
     <PackageVersion Include="Mono.Unix" Version="7.1.0-final.1.21458.1" />
-    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3"/>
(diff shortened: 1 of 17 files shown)


DeagleGross:
idea by Brennan: add the p90 collection

sebastienros (approved):

PR #792, opened 2025-02-07
feat: httpclient allow custom headers

DeagleGross:


src/Microsoft.Crank.Jobs.HttpClient/httpclient.yml
```diff
@@ -29,4 +29,5 @@ jobs:
...
       certPwd: # the password of the certificate specified in certPath
       sslProtocol: # the SSL protocol to use, e.g. Tls12, Tls13.
-    arguments: '--url "{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}" --connections {{connections}} --warmup {{warmup}} --duration {{duration}} {{headers[presetHeaders]}} --version {{httpVersion}} {% if certPath %} --cert {{certPath}} {% endif %} {% if certPwd %} --certpwd {{certPwd}} {% endif %} {% if sslProtocol %} --ssl-protocol {{sslProtocol}} {% endif %}'
+      customHeaders: [ ] # list of headers with the format: '<name1>: <value1>', e.g. [ 'content-type: application/json' ]. In yml should look like: - "content-type: application/json"
+    arguments: '--url "{{serverScheme}}://{{serverAddress}}:{{serverPort}}{{path}}" --connections {{connections}} --warmup {{warmup}} --duration {{duration}} {{headers[presetHeaders]}} {% for h in customHeaders %}{% assign s = h | split : ':' %}--header \"{{ s[0] }}: {{ s[1] | strip }}\" {% endfor %} --version {{httpVersion}} {% if certPath %} --cert {{certPath}} {% endif %} {% if certPwd %} --certpwd {{certPwd}} {% endif %} {% if sslProtocol %} --ssl-protocol {{sslProtocol}} {% endif %}'
```
DeagleGross, line 33: I copied the same code from bombardier.yml `{% for h in customHeaders %}{% assign s = h | split : ':' %}--header \"{{ s[0] }}: {{ s[1] | strip }}\" {% endfor %}`, but it reports such error in build-time. 
> Nested mappings are not allowed in compact mappings`

@sebastienros  can you please explain? I cant understand that

sebastienros: That's a yml issue in this case. In yml properties are string unless they start by a number, are `true, false`, json tokens `{, [``, or `-` for arrays.

In bombardier, since this property starts with `--` we force yaml to treat it as a string by enclosing the property in double-quotes `"`. So in bombardier to inject the `"` around the header argument we escape it with `\" {{header...}} \"`.

In this current file it happens that the property is enclosed in single-quotes `'`, so if you copy the escape sequence I don't know how it will behave. But then you don't need to escape the `"` around the header argument. Check how it's done for `--url`.

DeagleGross: changed to `"` to wrap the `arguments:` - seems to be an easier solution.

it works:
> crank --config https://github.com/aspnet/Benchmarks/blob/main/scenarios/rejection.benchmarks.yml?raw=true --config https://github.com/DeagleGross/crank/blob/dmkorolev/headers-for-clients/src/
Microsoft.Crank.Jobs.HttpClient/httpclient.yml?raw=true --scenario httpsys-header-symbols --profile aspnet-perf-win --load.job httpclient

sebastienros (approved):

PR #793, opened 2025-02-07
feat: update crank-agents to install latest openssl

DeagleGross:
verified by running `./build.sh` on wsl and then ran the container. Unfortunately, it is not available in the main distribution (maybe i did not find it correctly http://ftp.debian.org/debian/).
image

If you want, I can change base image from `mcr.microsoft.com/dotnet/sdk:8.0` to `mcr.microsoft.com/dotnet/sdk:8.0-alpine` like is done in [TLS benchmarks](https://github.com/aspnet/Benchmarks/blob/3773f786bb25e1f344370be398a91bf236cf3f07/src/BenchmarksApps/TLS/Kestrel/Dockerfile#L2) which provides easier way to install latest openssl version.


sebastienros (approved):
Merge on Monday in case it breaks anything this WE.

PR #795, opened 2025-02-10
Create Crank Release Yaml

LoopedBard3:
Creates a release yaml for Crank. This is the first pass using the automated template provided to use by 1ES. The tasks mostly match what we had in the previous manual release steps with the primary difference being a dryrun option and the path of the PackageArtifacts. The current most pressing question is whether the pool is setup properly and to run a test dry-run.


DrewScoggins (approved):
LGTM

sebastienros (approved):

PR #798, opened 2025-02-13
Clean-up and finish the release yaml

LoopedBard3:
Clean-up and finish the release yaml. This consisted of properly setting the pool to use, updating how the NuGetApiKey is passed in, and putting the publish step itself behind having to be run from main.

Test run with the dryrun features still included, but that shows variables are set properly and packages loaded: https://dev.azure.com/dnceng/internal/_build/results?buildId=2641796&view=results

Final test run with the main branch filter: https://dev.azure.com/dnceng/internal/_build/results?buildId=2642486&view=results. It correctly did not do the actual publish.


sebastienros (approved):

LoopedBard3:
I am not sure if there is a better way to gatekeep the running of the publish without causing errors in general, so if someone has thoughts, please share.

PR #832, opened 2025-08-20
Update package versions

sebastienros:


LoopedBard3 (changes requested):
Think there was one missed change, otherwise looks good.

LoopedBard3 (approved):
Looks good 👍.

PR #833, opened 2025-08-21
Detect powershell installation

sebastienros:


EgorBo (approved):
Just checked - works as expected on a system where `poweshell.exe` is not presented (only pwsh)

PR #834, opened 2025-08-21
Support job files in Azure Dev Ops Worker

LoopedBard3:
This adds support for sending files over Azure Service Bus to the Azure Dev Ops Worker. Changes include adding a <string, string> dictionary for the files to the JobPayload, added the default usage of a temp directory to use as the working directory for these files (including cleanup), and added functionality to materialize base64 files from the JobPayload files dictionary to actual files.

### DIFF
--- a/src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
+++ b/src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
@@ -210,8 +210,9 @@ namespace Microsoft.Crank.AzureDevOpsWorker
                             Console.WriteLine($"{LogNow} Job failed, attempt ({retries + 1} out of {jobPayload.Retries + 1}).");
                         }
 
-                        // The directory where the crank executable is located, taken from the default in the below crank job creation.
-                        var workingDirectory = Path.GetDirectoryName("crank");
+                        // Create a per-attempt temp working directory
+                        var workingDirectory = CreateTempWorkingDirectory();
+                        Console.WriteLine($"{LogNow} Created temp working directory: {workingDirectory}");
 
                         // Save the job payload files to disk (extracted step)
                         MaterializeFiles(jobPayload, workingDirectory);
@@ -271,6 +272,9 @@ namespace Microsoft.Crank.AzureDevOpsWorker
                                 await Task.Delay(TaskLogFeedDelay);
                             }
                         }
+
+                        // Attempt-specific cleanup of the temp working directory
+                        TryDeleteDirectory(workingDirectory);
 
                         retries++;
                     }
@@ -366,6 +370,9 @@ namespace Microsoft.Crank.AzureDevOpsWorker
                 ? Directory.GetCurrentDirectory()
                 : workingDirectory);
 
+            // Ensure base working directory exists
+            Directory.CreateDirectory(baseDir);
+
             foreach (var file in jobPayload.Files)
             {
                 var relativePath = file.Key?.Replace('\\', '/');
@@ -405,6 +412,30 @@ namespace Microsoft.Crank.AzureDevOpsWorker
                 {
                     Console.WriteLine($"{LogNow} Invalid base64 content for file: {relativePath}. Skipping. Content: {file.Value} \n\n");
                 }
+            }
+        }
+
+        // Create a unique temp working directory
+        private static string CreateTempWorkingDirectory()
+        {
+            var dir = Path.Combine(Path.GetTempPath(), "crank-" + Guid.NewGuid().ToString("N"));
+            Directory.CreateDirectory(dir);
+            return dir;
+        }
+
+        // Best-effort deletion of the temp directory
+        private static void TryDeleteDirectory(string dir)
+        {
+            try
+            {
+                if (!string.IsNullOrWhiteSpace(dir) && Directory.Exists(dir))
+                {
+                    Directory.Delete(dir, recursive: true);
+                }
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"{LogNow} Failed to delete temp directory '{dir}': {ex.Message}");
             }
         }
     }
--- a/src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
+++ b/src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
@@ -373,7 +373,7 @@ namespace Microsoft.Crank.AzureDevOpsWorker
 
             foreach (var file in jobPayload.Files)
             {
-                var relativePath = file.Key?.Replace('\\', '/');
+                var relativePath = file.Key?.Replace('\\', Path.DirectorySeparatorChar);
                 if (string.IsNullOrWhiteSpace(relativePath))
                 {
                     Console.WriteLine($"{LogNow} Skipping empty file path entry.");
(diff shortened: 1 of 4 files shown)


src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -133,7 +134,10 @@ private static async Task MessageHandler(ProcessMessageEventArgs args)
 
                 if (Verbose)
                 {
-                    Console.WriteLine($"{LogNow} Payload (Base64): {Convert.ToBase64String(bodyArray)}");
+                    // Truncate verbose payload logging to avoid huge console output
+                    var payloadB64 = Convert.ToBase64String(bodyArray);
+                    var max = 2048;
```
LoopedBard3, line 139: Is this truncation something we want? I think it makes sense to keep logging size from growing too big.

test/Microsoft.Crank.IntegrationTests/AzureWorkerTests.cs
```diff
@@ -72,5 +76,131 @@ public void ShouldParseTimeout(string timeSpan, string hexPayload)
...
+                    {
+                        ["nested/dir/a.txt"] = Convert.ToBase64String(Encoding.UTF8.GetBytes("A")),
+                        ["b.json"] = Convert.ToBase64String(Encoding.UTF8.GetBytes("{\"ok\":true}")),
+                    }
+                };
+
+                var method = typeof(Program).GetMethod("MaterializeFiles", BindingFlags.NonPublic | BindingFlags.Static);
+                Assert.NotNull(method);
+
+                method.Invoke(null, new object[] { payload, tmp });
```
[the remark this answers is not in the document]
LoopedBard3: Good Idea, fixed.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -344,5 +358,85 @@ private static Task ErrorHandler(ProcessErrorEventArgs args)
...
+                }
+
+                try
+                {
+                    var fileContent = Convert.FromBase64String(file.Value ?? string.Empty);
+                    File.WriteAllBytes(targetPath, fileContent);
+                }
+                catch (FormatException)
+                {
+                    Console.WriteLine($"{LogNow} Invalid base64 content for file: {relativePath}. Skipping. Content: {file.Value} \n\n");
```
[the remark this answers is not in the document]
sebastienros: Same as copilot. We should not log any sensitive information.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -344,5 +358,85 @@ private static Task ErrorHandler(ProcessErrorEventArgs args)
...
+                }
+            }
+        }
+
+        // Create a unique temp working directory
+        private static string CreateTempWorkingDirectory()
+        {
+            var dir = Path.Combine(Path.GetTempPath(), "crank-" + Guid.NewGuid().ToString("N"));
```
sebastienros, line 421: Prefer `Directory.Createtempsubdirectory("crank-
)` because it will set the correct permissions so only the current account can access it. The tmp folder on linux is visible from everyone.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
sebastienros: You can display the content size instead of the content

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -344,5 +356,85 @@ private static Task ErrorHandler(ProcessErrorEventArgs args)
...
+            var baseDir = Path.GetFullPath(string.IsNullOrEmpty(workingDirectory)
+                ? Directory.GetCurrentDirectory()
+                : workingDirectory);
+
+            // Ensure base working directory exists
+            Directory.CreateDirectory(baseDir);
+
+            foreach (var file in jobPayload.Files)
+            {
+                var relativePath = file.Key?.Replace('\\', '/');
```
DrewScoggins, line 376: I assume this is to make it work on Windows and Linux as both can use `/`, but should we be using `Path.Seperator` instead?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
```diff
@@ -344,5 +356,85 @@ private static Task ErrorHandler(ProcessErrorEventArgs args)
...
+                {
+                    Console.WriteLine($"{LogNow} Skipping empty file path entry.");
+                    continue;
+                }
+
+                var targetPath = Path.GetFullPath(Path.Combine(baseDir, relativePath));
+
+                // Ensure targetPath stays within baseDir to prevent path traversal
+                var normalizedBaseDir = Path.GetFullPath(baseDir).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+                var normalizedTarget = Path.GetFullPath(targetPath).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
```
DrewScoggins, line 387: nit: We don't need to call `Path.GetFullPath` again here

test/Microsoft.Crank.IntegrationTests/AzureWorkerTests.cs
```diff
@@ -72,5 +75,125 @@ public void ShouldParseTimeout(string timeSpan, string hexPayload)
 
             Assert.Equal(timeSpan, payload.Timeout.ToString());
         }
+
+        [Fact]
```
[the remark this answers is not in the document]
LoopedBard3: Yup, updated the tests with this alternative approach.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
sebastienros: Why resolved?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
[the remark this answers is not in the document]
sebastienros: Why resolved?

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
sebastienros: Why resolved?

LoopedBard3: Updated the log to only log the size of the payload.

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
LoopedBard3: Removed the creation of the temp directory and instead create the crank-* directory from the running directory of the devops worker.

sebastienros: I meant it literally, `Directory.CreateTempSubdirectory`

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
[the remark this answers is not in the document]
sebastienros: What does `Content: {file.Value}` contain?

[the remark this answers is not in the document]
LoopedBard3: The Content: {file.Value} has been removed as it would have been the string of the file. (https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.AzureDevOpsWorker/Program.cs#L410).

src/Microsoft.Crank.AzureDevOpsWorker/Program.cs
LoopedBard3: Ah, that makes sense. Do you have a preference on using the System temp with `Directory.CreateTempSubdirectory` or managing our own temp directories?

PR #841, opened 2025-08-28
Add Archive source option

sebastienros:
Ability to upload a ZIP file or have the agent pull one from a URL. This allows setup without the git client.


test/Microsoft.Crank.IntegrationTests/GitClientIntegrationTests.cs
LoopedBard3, line 1: This seems to be an extra or unsaved file.

src/Microsoft.Crank.Models/Source.cs
```diff
@@ -17,6 +17,7 @@ public class Source
         public string Repository { get; set; }
         public bool InitSubmodules { get; set; }
         public string LocalFolder { get; set; }
+    public string Archive { get; set; }
```
LoopedBard3, line 20: nit: spacing

PR #844, opened 2025-09-15
Add initScript to support authentication before trying docker 

LoopedBard3:
Add initScript to support authentication before trying docker source flow.

### DIFF
--- a/docker/agent/Dockerfile
+++ b/docker/agent/Dockerfile
@@ -88,22 +88,6 @@ RUN cd /tmp \
     && cp docker/docker /usr/bin \
     && rm -rf docker.tgz docker
 
-# Install Azure CLI - https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt#option-2-step-by-step-installation-instructions
-# We cannot use the helper script to install the CLI with one command as the script
-# gets an amd64-only DEB package while this image needs to support both amd64 and arm64.
-RUN apt-get update \
-    && apt-get install -y \
-        apt-transport-https \
-        ca-certificates \
-        lsb-release \
-        gnupg \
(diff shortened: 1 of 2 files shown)


src/Microsoft.Crank.Models/Job.cs
```diff
@@ -233,6 +233,7 @@ public Source Source
         public RollingLog Output { get; set; } = new RollingLog(10000);
         public bool SelfContained { get; set; }
         public string BeforeScript { get; set; }
+        public string DockerInitScript { get; set; }
```
sebastienros, line 236: Why not `InitScript` and also use it for non-docker path

LoopedBard3: Wasn't sure what level we would want this at but InitScript makes sense. Made those changes.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3304,6 +3323,26 @@ await ProcessUtil.RetryOnExceptionAsync(3, async () =>
...
+            if (!string.IsNullOrEmpty(job.InitScript))
+            {
+                try
+                {
+                    var environmentVariables = new Dictionary<string, string>()
+                    {
+                        ["CRANK_WORKING_DIRECTORY"] = benchmarkedApp
+                    };
+
+                    var segments = job.InitScript.Split(' ', 2);
```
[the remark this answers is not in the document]
caaavik-msft: After discussion with Parker, going to resolve this as this is reusing similar patterns from the other arguments like beforeScript etc. If someone wants to use this and has an executable on a path with a space in it, then they should put the command inside another script and invoke that or maybe using "bash -c ..." commands

PR #852, opened 2025-10-17
configure openssl setup on crank-agent Dockerfile

DeagleGross:
we need to test behavior of openssl with specific CipherSuites / ECs, so configured Dockerfile with `ENABLE_FIPS_MODE` build-arg to change openssl.

tested by re-running crank with `--enable-fips` flag, and then collected network dump where i see P384 EC instead of X25519:
image

### DIFF
--- a/docs/setup_linux.md
+++ b/docs/setup_linux.md
@@ -66,7 +66,8 @@ In order to restart and update the agent regularly, the following cron job can b
 crontab -e
 ```
 
-_Note: build.sh has parameters, so depending on the machine you could choose behavior building docker image:_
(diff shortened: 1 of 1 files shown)


sebastienros (approved):

sebastienros:
At some point we should consider that more changes that are really for a single scenario (or customer) should be in a dedicate space, like `aspnet/benchmarks` for us, and the machine pulls this instead of the one from crank. Since this feature is triggered by a flag that should be ok but we should not abuse it or this will become a "Crank" maintenant issue instead of a customer's one.

PR #873, opened 2026-03-18
Enable CFSClean* policies for dotnet-crank-ci-official pipeline

mmitche:
Adds CFSClean and CFSClean2 network isolation policies.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Crank.Agent
         private static readonly string _dotnetInstallShUrl = "https://dot.net/v1/dotnet-install.sh";
         private static readonly string _dotnetInstallPs1Url = "https://dot.net/v1/dotnet-install.ps1";
         private static readonly string _perfviewUrl = $"https://github.com/Microsoft/perfview/releases/download/{PerfViewVersion}/PerfView.exe";
-        private static readonly string _ultraUrl = $"https://www.nuget.org/api/v2/package/ultra/{UltraVersion}";
+        private static readonly string _ultraUrl = $"https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/flat2/ultra/{UltraVersion}/ultra.{UltraVersion}.nupkg";
 
         private static readonly string _aspnet8FlatContainerUrl = "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/flat2/Microsoft.AspNetCore.App.Runtime.linux-x64/index.json";
         private static readonly string _aspnet9FlatContainerUrl = "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/flat2/Microsoft.AspNetCore.App.Runtime.linux-x64/index.json";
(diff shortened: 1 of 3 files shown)


mmitche:
✅ Test build passed after fixing nuget.org references for CFSClean compliance.

**Passing build:** https://dev.azure.com/dnceng/internal/_build/results?buildId=2929815

**Changes in this update:**
- Replaced nuget.org URLs with dotnet-public AzDO feed equivalents for ultra package download and version checker
- Added \<clear />\ to NuGet.config files generated by the crank agent to prevent inheriting nuget.org from machine/user-level configs
- This ensures the crank agent and integration tests work correctly under CFSClean network isolation

src/Microsoft.Crank.Controller/VersionChecker.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Crank.Controller
     public static class VersionChecker
     {
         static TimeSpan CacheTimeout = TimeSpan.FromDays(1);
-        static string PackageVersionUrl = "https://api.nuget.org/v3-flatcontainer/microsoft.crank.controller/index.json";
+        static string PackageVersionUrl = "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/flat2/microsoft.crank.controller/index.json";
```
sebastienros, line 19: The dotnet tool which is installed from nuget.org is checking the nuget feed for new versions in order to display a warning. Should we care here?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6376,6 +6384,7 @@ public static async Task EnsureDotnetInstallExistsAsync()
                 File.WriteAllText(rootNugetConfig, @"<?xml version=""1.0"" encoding=""utf-8""?>
 <configuration>
   <packageSources>
+    <clear />
```
sebastienros, line 6387: We should be fine since this is added to the root folder

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -68,7 +68,7 @@ public class Startup
         private static readonly string _dotnetInstallShUrl = "https://dot.net/v1/dotnet-install.sh";
         private static readonly string _dotnetInstallPs1Url = "https://dot.net/v1/dotnet-install.ps1";
         private static readonly string _perfviewUrl = $"https://github.com/Microsoft/perfview/releases/download/{PerfViewVersion}/PerfView.exe";
-        private static readonly string _ultraUrl = $"https://www.nuget.org/api/v2/package/ultra/{UltraVersion}";
+        private static readonly string _ultraUrl = $"https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/flat2/ultra/{UltraVersion}/ultra.{UltraVersion}.nupkg";
```
sebastienros, line 71: @LoopedBard3 we need to update the package version and push it in the internal feed

src/Microsoft.Crank.Controller/VersionChecker.cs
mmitche: If it's installed from nuget.org...wouldn't there not ever be a newer package anyway?

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Pushed v1.3.0 and v2.0.2 to the feed. We can merge this as is and then update to v2.0.2 if we want.

LoopedBard3 (approved):
Looks good to me.

PR #874, opened 2026-03-20
Strip version metadata from Results dict when `--no-metadata` is used

ilonatommy:
## Problem

  The ASP.NET Core benchmarks Power BI dashboard (KPIs page) shows two broken rows: "NETCore App" and "ASP.NET Core" that display raw version strings (e.g. 10.0.5+a612c2a1056f) instead of KPI charts
I was convinced some benchmarks are broken and was looking for failed jobs to fix them, it does not look intentional:

image

  These are not real benchmark scenarios. No scenario YAML in aspnet/Benchmarks (https://github.com/aspnet/Benchmarks) defines them. They are phantom rows created when Power BI reads text-valued version entries from the results dictionary in stored JSON.

## Root cause

  The crank Agent registers AspNetCoreVersion and NetCoreAppVersion as measurements (Startup.cs (https://github.com/dotnet/crank/blob/main/src/Microsoft.Crank.Agent/Startup.cs#L3238-L3301)). The Controller's AggregateAndReduceResults() aggregates all measurements — including these text strings — into jobResult.Results.

  The baseline pipelines pass `--no-metadata --no-measurements`, but:
   - `--no-metadata` only clears Metadata[] (the description array)
   - `--no-measurements` only clears Measurements[] (the raw samples)
   - Neither touches Results — the aggregated values persist in the stored JSON

  This has been the case since the initial commit (0347e98, Jul 2020).

  ## Fix

  When `--no-metadata` is active, also remove the well-known version keys from `jobResult.Results`:
   - netCoreAppVersion / NetCoreAppVersion
   - aspNetCoreVersion / AspNetCoreVersion
   - netSdkVersion

### DIFF
--- a/src/Microsoft.Crank.Controller/Program.cs
+++ b/src/Microsoft.Crank.Controller/Program.cs
@@ -2890,6 +2890,18 @@ namespace Microsoft.Crank.Controller
             }
         }
 
+        // Version result keys that are text metadata, not numeric benchmarks.
+        // When stored alongside numeric results they pollute data consumers
+        // (e.g. Power BI) that expect every result key to be a chartable metric.
+        private static readonly string[] VersionResultKeys = new[]
+        {
+            "netCoreAppVersion",
+            "NetCoreAppVersion",
+            "aspNetCoreVersion",
+            "AspNetCoreVersion",
+            "netSdkVersion",
+        };
+
         private static void CleanMeasurements(JobResults jobResults)
         {
             // Remove metadata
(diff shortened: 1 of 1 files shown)


sebastienros:
> shows two broken rows: "NETCore App" and "ASP.NET Core" that display raw version strings (e.g. 10.0.5+a612c2a1056f) instead of KPI charts

I am confused, because they do exactly what they are meant for ;) They are placed as-is by binding the specific fields to Text outputs so we can see what version the KPIs were run with.

sebastienros:
Also these are not "metadata" the way I see it. It's a real value of a run, hence why it's put in results directly. There are ways to alter the returned document if you don't want specific results, or want to add new ones. See how the 90th percentile of CPU is computed dynamically.

ilonatommy:
> I am confused, because they do exactly what they are meant for ;) They are placed as-is by binding the specific fields to Text outputs so we can see what version the KPIs were run with.

Honestly, I spent quite a lot of time investigating why there's no data displayed for the last 2 rows even thought the machines the jobs are running on (visible when you hover over) are green. I had no idea that's meant to look like this. Closing.

PR #878, opened 2026-04-14
Add `ci` channel for per-commit Build Cache Service framework resolution

LoopedBard3:
Adds a new `ci` channel that resolves .NET binaries from the Build Cache Service (BCS) instead of NuGet feeds, giving **per-commit** granularity for performance-regression bisection instead of feed cadence.

On the `ci` channel, crank **always overlays BOTH shared frameworks** in every job — the base runtime (`Microsoft.NETCore.App`, from dotnet/runtime) **and** the ASP.NET Core shared framework (`Microsoft.AspNetCore.App`, from dotnet/aspnetcore). Both default to **latest**, and each repo can be pinned independently to bisect one while holding the other at latest.

> Named `ci` (not `buildcache`) per review feedback — "buildcache" is an implementation detail; `ci` describes where the binaries come from, consistent with the other channel names.

## CLI surface

Rather than add new arguments, the `ci` channel **reuses the existing `runtimeVersion` / `aspNetCoreVersion` arguments** — on this channel they carry a **commit SHA** instead of a feed version:

- **empty** → latest cached build for that repo on `main`
- **8–40 hex chars** → pin/bisect that repo at that commit
- **a feed version string** → **rejected with a clear error** (version-string pinning is not supported on `ci`; conversely the other channels accept a version string and reject a SHA)

No new arguments are introduced. *Latest* always resolves against `main` — the only branch the BCS pipeline builds — so there's no branch knob to configure. To target a specific point, pin a commit SHA; SHA lookups are branch-independent, so a commit that happens to live on a release branch still resolves. (If the pipeline starts caching release branches later, branch selection can be added back as a purely additive argument.)

**Usage:**
```
# latest of both on main
--application.channel ci

# bisect aspnetcore, runtime stays latest
--application.channel ci --application.aspNetCoreVersion <aspnetcore-sha>

# bisect the base runtime, aspnetcore stays latest
--application.channel ci --application.runtimeVersion <runtime-sha>

# pin both
--application.channel ci --application.runtimeVersion <runtime-sha> --application.aspNetCoreVersion <aspnetcore-sha>
```

**Controller (`Job.cs`)** — reuses `RuntimeVersion` / `AspNetCoreVersion` (no new properties); on the `ci` channel they carry a commit SHA.

**Agent (`Startup.cs` / `BuildCacheClient.cs`)** — new CLI:
- `--build-cache-base-url` (default: `https://pvscmdupload.z22.web.core.windows.net`)
- `--build-cache-disabled`

## How resolution works

The agent still builds the app with real NuGet packages (so `restore`/`publish` behave normally), then overlays BCS bits per publish kind:
- **Framework-dependent**: a per-job isolated dotnet home is created at `<temp>/crank-buildcache/home-…/`, mirrored from the global `_dotnethome`, and overlaid with the BCS runtime **and** aspnetcore archives. `StartProcess` resolves from this home; the global `_dotnethome` is never mutated (concurrent jobs / other channels unaffected).
- **Self-contained**: BCS files are overlaid onto the published `outputFolder`; the SDK-bound apphost is preserved.

The base runtime archive is raw build output, so BCS bits are overlaid onto a feed-installed runtime. The aspnetcore archive is the runtime-pack nupkg stored verbatim (carrying `deps.json` + `runtimeconfig.json`), so that framework folder is built entirely from BCS and the job **fails loudly** if the pack is incomplete. If a requested commit isn't in the cache, crank fails rather than silently falling back.

## Version surfaces (bisection distinguishability)

Every `ci` run stamps the resolved commits so bisection points never collide, and the stamp reaches the perf database:
- results measurements `NetCoreAppVersion` / `AspNetCoreVersion` (what lands in the controller `--json` and the SQL/ES document as `application.netCoreAppVersion` / `aspNetCoreVersion`) → `{feedVersion}+ci.{12-char sha}` on the `ci` channel; non-`ci` runs are **unchanged** (`{feedVersion}+{12-char sha}`). The `+ci.` marker is folded into the existing fields rather than added as new keys, so bisection runs are self-identifying without any downstream schema change.
- `job.RuntimeVersion` / `job.AspNetCoreVersion` (the DTO, visible on the agent job object) → `{feedVersion}+ci.{8-char sha}`
- `Dependency` records carry the full 40-char `CommitHash` + repo URL for both runtime and aspnetcore

The `12`- and `8`-char forms and the full `Dependency` hash all encode the same underlying commit; the numeric version prefix stays feed-coherent (so runtimeconfig/roll-forward stay valid) and the `+ci.{sha}` suffix is the per-commit distinguisher.

## Cache correctness

`RuntimeVersion` and `AspNetCoreVersion` are part of `BuildKeyData`, so `--application.options.reuseBuild` cache-busts when a pinned SHA changes. The reuse path also records the `+ci.{sha}` version measurements (it returns before the fresh-build recording, so without this a reused `ci` run would surface no version).

## Backward compatibility

Fully gated on `--application.channel ci`. `DefaultChannel` is unchanged (`current`). No new controller properties are added — the existing `runtimeVersion` / `aspNetCoreVersion` arguments only change meaning **on the `ci` channel**; every other channel takes the exact same path as before, so old controllers/agents interoperate unchanged. Audited diff-level against `main`: purely additive, no opt-out-path behavior change.

## Verification

- `dotnet build` clean; **119/119** unit tests pass (`BuildCacheClientTests.cs` adds 40 methods covering `ParseLatestBuilds`, commit-SHA validation incl. path-traversal, the SHA-vs-version channel rules, dual-repo overlay, isolated-home concurrency, `SelectHighestManagedDir` numeric ordering, cleanup, etc.).
- Local both-overlay hello-world runs (win-x64) confirmed the three `ci` paths end-to-end:
  - **latest** resolution engages the correct `latest/main/latestBuilds.json` path per repo;
  - **pinned aspnetcore + latest runtime** overlays both and runs, with results surfacing `aspNetCoreVersion = {ver}+ci.8f51f037469f` and `netCoreAppVersion = {ver}+ci.6550ccf7827c` (matching the resolved commits), while the non-`ci` load job carries no `ci.` marker (gating intact);
  - a **feed version string** on `ci` fails fast with the documented error.

> **Post-merge dependency:** the aspnetcore `latestBuilds.json` index (published by the BCS indexer / dotnet-performance pipeline) must exist for latest-aspnetcore resolution; until then the agent fails loudly rather than silently falling back.


src/Microsoft.Crank.Agent/BuildCacheClient.cs
```diff
@@ -0,0 +1,781 @@
...
+
+        private static string ExtractVersionPrefix(string targetFramework)
+        {
+            // "net10.0" → "10.0", "net9.0" → "9.0"
+            if (targetFramework != null && targetFramework.StartsWith("net", StringComparison.OrdinalIgnoreCase))
+            {
+                return targetFramework.Substring(3);
+            }
+
+            return "10.0";
```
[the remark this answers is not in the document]
LoopedBard3: Obsoleted by 1ecc2393 / 19f909e4: `ExtractVersionPrefix` was part of the dead `OverlayRuntimeAsync` / `InstallRuntimeFromBuildCacheAsync` design and was removed. Round 3 introduced `SelectHighestManagedDir` which scans the actual `shared/<framework>/` directories produced by the BCS archive and picks the highest using numeric-aware comparison (`net11.0` > `net10.0` > `net9.0`), so no TFM parsing is needed.

src/Microsoft.Crank.Agent/BuildCacheClient.cs
```diff
@@ -0,0 +1,781 @@
...
+                Log.Info($"Build Cache: Copied {filesCopied} managed assemblies to shared framework.");
+            }
+
+            // Copy native libraries
+            if (nativeDir != null && Directory.Exists(nativeDir))
+            {
+                int nativeCount = 0;
+                var nativeExtensions = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                    ? new[] { "*.dll", "*.exe" }
+                    : new[] { "*.so", "*.dylib", "*" };
```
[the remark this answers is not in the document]
LoopedBard3: Obsoleted by 1ecc2393 (round 1): `nativeExtensions` was inside `OverlayRuntimeAsync`, which was deleted as part of the dead-code cleanup. The current `CopyNative` in `BuildCacheClient.cs` filters explicitly by extension/symbol pattern and doesn't have an unused local.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3434,6 +3505,24 @@ await ProcessUtil.RetryOnExceptionAsync(3, async () =>
...
+                {
+                    try
+                    {
+                        var overlayCount = BuildCacheClient.OverlayPublishedOutput(buildCacheExtractDir, outputFolder);
+                        Log.Info($"Build Cache: Overlaid {overlayCount} runtime files from commit {buildCacheCommitSha.Substring(0, 8)} into published output");
+
+                        // Update the reported runtime version to reflect the BCS commit
+                        runtimeVersion = $"{runtimeVersion}+buildcache.{buildCacheCommitSha.Substring(0, 8)}";
+                    }
+                    catch (Exception ex)
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 1ecc2393 (round 1) and 56657e2a (round 2): the BCS commit is no longer appended to `runtimeVersion`. `PatchRuntimeConfig` keeps the feed-resolved version (which is actually installed), and the BCS commit is reported separately via `MeasurementMetadata` (`runtime-version`). FDD startup is no longer broken by the rewrite.

src/Microsoft.Crank.Agent/BuildCacheClient.cs
```diff
@@ -0,0 +1,781 @@
...
+            {
+                Log.Info($"Build Cache: Using cached archive at {tempArchive}");
+            }
+
+            var extractDir = Path.Combine(tempDir, "extracted");
+            if (Directory.Exists(extractDir))
+            {
+                Directory.Delete(extractDir, true);
+            }
+
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 19f909e4 (round 3): per-call extract directories now use `extracted-{config}-{guid}` naming, and `CleanupExtractDir` removes them in `DeleteJobAsync`'s `finally` block. The per-job isolated dotnet home (`home-<shortsha>-<config>-<guid>`) is also cleaned up. Downloaded archives are still cached under `<sha>/<config>-<filename>` for reuse across jobs.

src/Microsoft.Crank.Agent/BuildCacheClient.cs
```diff
@@ -0,0 +1,781 @@
...
+                // Try to get the config-specific entry, fall back to "all"
+                if (latestBuilds.Entries.TryGetValue(buildCacheConfig, out var configEntry) && !string.IsNullOrEmpty(configEntry.CommitSha))
+                {
+                    commitSha = configEntry.CommitSha;
+                    Log.Info($"Build Cache: Using latest commit {commitSha.Substring(0, 8)} for config '{buildCacheConfig}' on branch '{branch}' (committed {configEntry.CommitTime})");
+                }
+                else if (latestBuilds.Entries.TryGetValue("all", out var allEntry) && !string.IsNullOrEmpty(allEntry.CommitSha))
+                {
+                    commitSha = allEntry.CommitSha;
+                    Log.Info($"Build Cache: Using latest commit {commitSha.Substring(0, 8)} for all configs on branch '{branch}' (committed {allEntry.CommitTime})");
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 1ecc2393 (round 1): introduced a `ShortSha()` helper that uses `commitSha.Substring(0, Math.Min(8, commitSha.Length))`. All logging call sites go through it. Round 3 also added `ValidateCommitSha` to reject invalid SHAs at the boundary.

docs/build_cache_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+**Status:** Already in progress (per prior discussion).
+
+Crank's BCS client uses unauthenticated HTTP GET requests to download artifacts. The blobs in the `pvscmdupload` storage account's `$web` container need to be publicly readable.
+
+**URLs crank will hit:**
+
+```
+GET https://pvscmdupload.blob.core.windows.net/$web/builds/{repoName}/latest/{branch}/latestBuilds.json
+GET https://pvscmdupload.blob.core.windows.net/$web/builds/{repoName}/buildArtifacts/{commitSha}/{configKey}/{artifactFile}
+```
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 1ecc2393 (round 1): the doc URLs in `docs/build_cache_requirements.md` now match what the agent actually calls — the static-website endpoint `https://pvscmdupload.z22.web.core.windows.net/builds/...` (no `/$web/`). The default `--build-cache-base-url` value is documented and matches.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -2933,23 +2957,70 @@ private static async Task<string> CloneRestoreAndBuild(string path, Job job, str
...
+                    // Download and extract the BCS artifacts to a temp directory
+                    buildCacheExtractDir = await BuildCacheClient.DownloadAndExtractAsync(
+                        _buildCacheBaseUrl, _buildCacheRepoName, buildCacheCommitSha, buildCacheConfig,
+                        targetFramework, cancellationToken);
+
+                    Log.Info($"Build Cache: Artifacts for commit {buildCacheCommitSha.Substring(0, 8)} ready for post-build overlay");
+
+                    // Resolve a REAL runtime version from feeds for the NuGet build
+                    runtimeVersion = await ResolveRuntimeVersion(buildToolsPath, targetFramework, "Latest");
+                    Log.Info($"Runtime for build: {runtimeVersion} (Latest from feeds, will be overlaid with BCS commit {buildCacheCommitSha.Substring(0, 8)})");
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 19f909e4 (round 3): `ValidateCommitSha` now rejects SHAs shorter than 8 chars (and anything outside `[0-9a-fA-F]{8,40}`) up-front, so the short-SHA path is unreachable here. All call sites that log a short SHA also go through `ShortSha()` which uses `Math.Min` defensively.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -3434,6 +3505,24 @@ await ProcessUtil.RetryOnExceptionAsync(3, async () =>
...
+                        // Update the reported runtime version to reflect the BCS commit
+                        runtimeVersion = $"{runtimeVersion}+buildcache.{buildCacheCommitSha.Substring(0, 8)}";
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.Info($"Build Cache: Warning - overlay failed: {ex.Message}. Published app will use feed-sourced runtime.");
+                    }
+                }
+
                 PatchRuntimeConfig(job, outputFolder, aspNetCoreVersion, runtimeVersion);
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 19f909e4 (round 3) with a different approach: instead of mutating the global `_dotnethome` (which would create cross-job pollution), the agent now creates a per-job isolated dotnet home at `<temp>/crank-buildcache/home-<shortsha>-<config>-<guid>/`, mirrored from the global home and overlaid with BCS bits. `StartProcess` uses `jobContext.BuildCacheDotnetHome` for runtime resolution. FDD works correctly and concurrent jobs are isolated. The original `InstallRuntimeFromBuildCacheAsync` method was dead code and was deleted in round 1.

src/Microsoft.Crank.Agent/BuildCacheClient.cs
```diff
@@ -0,0 +1,781 @@
...
+    {
+        private static readonly HttpClient _httpClient = new HttpClient { Timeout = TimeSpan.FromMinutes(10) };
+
+        // Cache latestBuilds.json responses to avoid repeated downloads
+        private static readonly Dictionary<string, (DateTimeOffset fetchedAt, LatestBuildsResponse data)> _latestBuildsCache = new();
+        private static readonly TimeSpan _latestBuildsCacheDuration = TimeSpan.FromHours(1);
+
+        // Cache of already-installed BCS commit SHAs to avoid re-extracting
+        private static readonly HashSet<string> _installedBuildCacheRuntimes = new(StringComparer.OrdinalIgnoreCase);
+
```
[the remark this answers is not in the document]
LoopedBard3: Addressed by 1ecc2393 / 19f909e4: `_latestBuildsCache` is now a `ConcurrentDictionary` keyed by `$"{baseUrl}|{repoName}|{branch}"` so the base URL is included and concurrent reads/writes are safe. `_installedBuildCacheRuntimes` was part of the dead `InstallRuntimeFromBuildCacheAsync` design and was removed.

docs/dotnet_versions.md
```diff
@@ -55,9 +55,12 @@ When a TFM is configured, the agent will download the corresponding .NET SDK ver
...
 - `latest`: latest versions used by ASP.NET 
 - `edge`: latest nightly builds available
+- `buildcache`: base runtime + ASP.NET Core from the Build Cache Service (per-commit builds)
 
 The difference between `latest` and `edge` is that `latest` will pick runtimes and SDKs that are deemed compatible together. For instance a very recent .NET core runtime might be compatible with a less recent ASP.NET runtime. The `edge` is used to pick the absolute latest build for the select TFM.
 
+The `buildcache` channel uses the Build Cache Service (BCS) from `dotnet-performance-infra` to resolve framework versions by individual commit SHA rather than from VMR feeds. This provides much finer-grained control — every cached commit is available, whereas VMR feeds may have multi-day gaps between ingested commits. On this channel crank overrides **both** the base .NET runtime (`Microsoft.NETCore.App`, from dotnet/runtime) **and** the ASP.NET Core shared framework (`Microsoft.AspNetCore.App`, from dotnet/aspnetcore); each defaults to the latest cached build and can be pinned independently. SDK and desktop versions are resolved from `latest`.
```
sebastienros: I would prefer to name it `ci`, because buildcache sounds too much like an implementation detail.

docs/dotnet_versions.md
```diff
@@ -115,4 +118,71 @@ The following command uses the `edge` channel but ASP.NET is fixed so it doesn't
...
+### Basic usage (latest cached build of both frameworks on main)
+
+```
+> crank --config benchmarks.yml --scenario json --profile aspnet-perf-lin --application.channel buildcache
+```
+
+### Bisecting ASP.NET Core (pin aspnetcore, runtime stays latest)
+
+```
+> crank --config benchmarks.yml --scenario json --profile aspnet-perf-lin --application.channel buildcache --application.buildCacheAspNetCoreCommitSha a1b2c3d4e5f6...
```
sebastienros: Since we use a custom channel, which makes sense (what feed is the runtime coming from), would it be possible to reuse the exist "version" arguments instead of adding new ones? They will be shas and not actual versions.

And not setting a version would pick the latest one from the ci channel (buildcache)

sebastienros: I could see a world where we could set either a version or a sha, since I assume all builds actually have a version.

docs/dotnet_versions.md
```diff
@@ -115,4 +118,71 @@ The following command uses the `edge` channel but ASP.NET is fixed so it doesn't
...
+    --application.buildCacheRuntimeCommitSha 1111aaaa2222bbbb... \
+    --application.buildCacheAspNetCoreCommitSha 3333cccc4444dddd...
+```
+
+If a requested commit is not found in the cache, crank fails with an error rather than falling back.
+
+### Different branch (per repo)
+
+```
+> crank --config benchmarks.yml --scenario json --profile aspnet-perf-lin --application.channel buildcache \
```
sebastienros: Does it mean we can target a release branch and it will pick up the latest available build for this branch?

I was assuming it would only be for the `main` branches,

sebastienros: I don't see why this would be two different variables, no way we would use different branches for aspnet/dotnet. 

Another option would be to reuse the existing version arguments which already have the convention with patterns. Creating a custom format for wildcard versions.

docs/dotnet_versions.md
```diff
@@ -115,4 +118,71 @@ The following command uses the `edge` channel but ASP.NET is fixed so it doesn't
...
+### Build Cache properties
+
+| Property | Default | Description |
+|----------|---------|-------------|
+| `buildCacheRuntimeCommitSha` | (empty) | Specific [dotnet/runtime](https://github.com/dotnet/runtime) commit SHA to overlay onto `Microsoft.NETCore.App`. If empty, uses the latest cached runtime build for its branch. |
+| `buildCacheAspNetCoreCommitSha` | (empty) | Specific [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore) commit SHA to place as `Microsoft.AspNetCore.App`. If empty, uses the latest cached aspnetcore build for its branch. |
+| `buildCacheRuntimeBranch` | `main` | Branch to query for the latest runtime build. |
+| `buildCacheAspNetCoreBranch` | `main` | Branch to query for the latest aspnetcore build. |
+
+The BCS configuration key (e.g., `coreclr_x64_linux` for runtime, `aspnetcore_x64_linux` for aspnetcore) is auto-detected per repo from the agent platform. Platforms with no aspnetcore config (there is no macOS/musl/arm32 in v1) fail loud rather than silently skipping.
```
sebastienros, line 171: this is the kind of idea I am having when suggesting to use existing arguments, and base the logic on the channel.

docs/dotnet_versions.md
LoopedBard3: That makes sense, especially with the other channel names. Went ahead and updated this everywhere user-facing, but kept some of the internals as they truly refer to the buildCache.

docs/dotnet_versions.md
LoopedBard3: Moved over to this approach. The ci channel only supports shas because the Cache is keyed by commit and currently doesn't have any mapping or relation to the longer version strings. We would have to decide how we want to map versions to sha's as the VMR does not have a version for every ASP.NET or Runtime commit, and I am not aware of any other place that would have a version for every SHA (though maybe the mapping doesn't have to cover every SHA for general use).

docs/dotnet_versions.md
LoopedBard3: Landed on this IIUC.

docs/dotnet_versions.md
LoopedBard3: Yea, this was an idea of where this could be down the line with the idea that custom branches would be selectable from the cache for testing. Instead, I have removed ciBranch and this flow as a whole as we are currently only building for main so the functionality would not actually work for anything. If we find we need this in the future we can add it then, otherwise specific commits can be used from wherever they are built.

sebastienros (approved):

PR #886, opened 2026-06-08
Modernize crank's dotnet-trace plumbing (no default-behavior change)

LoopedBard3:
PR A of the dotnet-trace modernization series. **Strictly additive.** Every existing `--application.dotNetTrace true` invocation produces the same trace contents it did before, byte-for-byte.

## What changed

- **Package bump (minimal).** `Microsoft.Diagnostics.NETCore.Client` 0.2.621003 → 0.2.661903 to pick up the 3-arg `StartEventPipeSession(providers, requestRundown, circularBufferMB)` overload. **`Microsoft.Diagnostics.Tracing.TraceEvent` stayed at 3.1.23** — bumping it transitively forces `System.Text.Json` 8.0.5 → 9.0.8 and cascades into `System.Text.Encodings.Web` and `System.IO.Pipelines` 9.x across every project in the solution. That's a runtime-version-pack-style change that doesn't belong in a strictly additive PR. Crank uses TraceEvent only for stable `ClrTraceEventParser.Keywords` enum values, so the bump is safe to defer to its own PR.

- **`TraceExtensions.cs` refreshed from `dotnet/diagnostics`'s current `ProviderUtils.cs` and `ListProfilesCommandHandler.cs`:**
  - Modern CLR keywords added: `monitoring`, `codesymbols`, `eventsource`, `compilation`, `compilationdiagnostic`, `methoddiagnostic`, `typediagnostic`, `jitinstrumentationdata`, `profiler`, `waithandle`, `allocationsampling`.
  - Modern aliases added: `assemblyloader` (= `fusion`) and `managedheapcollect` (= `gcheapcollect`). Legacy keys preserved.
  - Corrected `gcsampledobjectallocation{high,low}` aliases added; legacy typo'd `gcsampledobjectallcation{high,low}` preserved.
  - Modern profiles added: `dotnet-common`, `dotnet-sampled-thread-time`, `sample-profiler`, `database`.
  - **The legacy `cpu-sampling` profile is intentionally left untouched** — that's the empty-providers fallback for `--dotNetTrace true`, and changing it would be a silent behavior change for every existing crank pipeline.

- **New `Job` knobs** (defaults reproduce today's exact behavior):
  - `DotNetTraceBufferSizeMB` — int, default 256 (today's hardcoded value). Surface: `--[JOB].dotnetTraceBufferSizeMB`.
[...]

PR #887, opened 2026-06-08
Add collect-linux mode to dotnet-trace integration

LoopedBard3:
## What

Adds a `collect-linux` mode to crank's `dotnet-trace` integration so Linux agents on modern kernels can capture whole-machine, perf_event-based traces with kernel + native frames, without LTTng.

New `Job` properties (defaults reproduce today's exact behavior):
- `DotNetTraceCollectMode` (`default` | `collect` | `collect-linux`) — `default` keeps today's in-process `DiagnosticsClient.StartEventPipeSession` path; `collect` and `collect-linux` shell out to the `dotnet-trace` CLI.
- `DotNetTraceStopTimeoutSec` — grace period after SIGINT before the CLI is killed. `0` selects the per-mode default (60s for `collect`, 180s for `collect-linux`).

Agent additions in `Startup.cs`:
- Pinned `dotnet-trace 10.0.721401` (latest on nuget.org with `collect-linux`; ships a `linux-arm64` build).
- Lazy `dotnet tool install --tool-path` on first CLI-mode use, with a post-install `--version` assert and a single-flight semaphore.
- `CollectViaDotNetTraceCliAsync` spawns the CLI directly (not via `ProcessUtil.RunAsync`) so we keep lifecycle control: wait for the benchmark to signal stop → send SIGINT → wait per-mode grace → fall back to `Kill(entireProcessTree:true)` on expiry.
- Grace-period expiry logs `[WRN] dotnet-trace did not finalize within {N}s; trace may be incomplete` and **keeps the partial trace** — does not fail the job.

Provider-string compatibility:
- `BuildDotnetTraceCliArgs` classifies each token into profile / provider-spec / CLR-keyword-expression using crank's existing `TraceExtensions` table.
- **Required rewrite**: `cpu-sampling` → `dotnet-sampled-thread-time` in `collect` mode only (the modern `cpu-sampling` profile has `VerbExclusivity="collect-linux"` and is rejected by the `collect` verb).
- **Cosmetic alias rewrites**: `fusion` → `assemblyloader`, `gcheapcollect` → `managedheapcollect`, plus the two `gcsampledobjectallcation{high,low}` typos.
- All rewrites are grouped into one `[INF] dotnet-trace providers normalized: …` line per job.
- Empty-providers default is `cpu-sampling` for `collect-linux` and `dotnet-sampled-thread-time,dotnet-common` for `collect`. **Empty-providers default for `default` mode is unchanged** (still `cpu-sampling` via PR A's preserved behavior).

`collect-linux` preflight (hard fail — **no silent fallback** to `collect`, per design):
- OS == Linux, kernel ≥ 6.4 (parsed from `/proc/sys/kernel/osrelease`), effective UID == 0 (via `Mono.Unix.Native.Syscall.geteuid()`).
- Verbatim error text:
  `DotNetTraceCollectMode=collect-linux requires Linux, root (effective UID 0), and kernel >= 6.4. Detected: OS={os}, EUID={euid}, kernel={ver}. Set DotNetTraceCollectMode=collect to use EventPipe collection instead.`

Docs:
- New flags documented in `Documentation.cs` + `README.md`.
- Decision matrix expanded from 2 rows (PR A) → 3 rows: `DotNetTrace=true` (any OS), `DotNetTrace=true` + `DotNetTraceCollectMode=collect-linux` (modern Linux + root), `Collect=true` (legacy PerfView pipelines).

## Why

LTTng-based `Collect=true` is structurally fragile on modern Ubuntu — cobalt-hosted Ubuntu 24.04 aarch64 returns empty managed-event sequences even with `lttng-tools` + `liblttng-ust-dev` installed correctly (likely runtime↔liblttng-ust ABI mismatch). `dotnet-trace collect-linux` is the perf_event-based replacement: same kernel-sampling capability as perfcollect provided, but without the LTTng dependency. Cobalt hosts (kernel 6.8, agent container runs as root) meet every prerequisite.

Existing `--application.dotNetTrace true` users keep their exact current behavior. `collect-linux` is purely opt-in via the new mode flag.

## Notes

- **`collect-linux` rejects `--buffersize`** (`TreatUnmatchedTokensAsErrors=true` in upstream's `CollectLinuxCommand`; perf_event uses kernel ring buffers, not EventPipe circular buffers). The argv builder strips it in that mode regardless of `DotNetTraceBufferSizeMB`.
- **Pinned tool version targets net8.0** (the tool's own TFM). `collect-linux` requires the **traced process** to be .NET 10+. If a user points it at a .NET 8/9 process, the CLI's stderr ("EventPipe IPC command not understood") is captured verbatim into `job.Error`.
- **Effective UID, not real UID.** `Mono.Unix.UnixUserInfo.RealUserId` would return the pre-sudo UID under elevation; `geteuid()` is the signal that matters for `perf_event_open` capability checks. (Rubber-duck blind-spot fix from the planning session.)
- **Windows `collect` mode is best-effort.** `GenerateConsoleCtrlEvent` doesn't reliably deliver Ctrl+C to non-console children spawned with `UseShellExecute=false`; the helper falls back to `CloseMainWindow()` + the grace-period kill. Windows users almost always want `default` mode anyway.

### Mixed-version compatibility

| Direction | Behavior |
|---|---|
| **New controller + old agent** | Old agent's deserializer ignores `DotNetTraceCollectMode` / `DotNetTraceStopTimeoutSec`. If the controller asks for `collect-linux`, the old agent silently runs `default`. Documented expectation. |
| **New agent + old controller** | New properties default to today's exact behavior (`default` mode, `0`-as-per-mode-default timeout). Byte-for-byte identical to today. |

Validated by `JobMixedVersionCompatTests` covering legacy and PR-A-era payloads.

## Tested

- `dotnet build Microsoft.Crank.sln -c Release` → clean (0 errors, 0 warnings).
- `dotnet test test/Microsoft.Crank.UnitTests` → **103 / 103 passing**, including 33 new tests across:
  - `DotNetTraceCliNormalizationTests` (BuildDotnetTraceCliArgs / NormalizeClrEventExpression — defaults per mode, cpu-sampling rewrite, classification, alias rewrites, buffer-size gating)
  - `DotNetTraceCollectLinuxPreflightTests` (ParseKernelVersion across the 6.4 boundary, error-message contract on non-Linux)
  - Extended `JobMixedVersionCompatTests` (PR B knob defaults + round-trip)
- ❌ **Manual `collect-linux` validation on cobalt aarch64 Ubuntu 24.04 host — pending reviewer.** Manual-validation checklist before un-drafting:
  - [ ] `dotnet tool install dotnet-trace --version 10.0.721401 --tool-path /tmp/x` resolves an **arm64** tool on the cobalt host (no x64-only fallback).
  - [ ] A simple .NET 10 benchmark with `--application.dotNetTrace true --application.dotNetTraceCollectMode collect-linux` produces a non-empty `.nettrace` with both managed and native frames.
  - [ ] Same invocation on a non-Linux/non-root environment fails fast with the documented preflight error string (no silent fallback to `collect`).

## Operator notes

dotnet-trace collect-linux reads tracefs to translate `perf_event_open` records. In a container the `/sys/kernel/tracing` directory exists but is empty unless the host's tracefs is bind-mounted, surfacing as `Error: Tracefs is not accessible: It appears tracefs is not mounted.` at runtime. This PR adds `-v /sys/kernel/tracing:/sys/kernel/tracing` to `docker/agent/run.sh` so the wrapper script handles this for new agent deployments. EventPipe-based collection (`DotNetTraceCollectMode=collect` or `Collect=true` perfcollect) does not need this mount; only `collect-linux` does. The host must also have tracefs mounted (Ubuntu 24.04 + systemd does this at boot; check with `mount | grep tracefs`). Operators upgrading an existing agent need to `./docker/agent/stop.sh && ./docker/agent/run.sh` to pick up the new mount -- volume mounts cannot be added to a running container. Validated on the cobalt-hosted aarch64 Ubuntu 24.04 (kernel 6.8) host.

The pinned `dotnet-trace` CLI version is **10.0.721401** (tag `v10.0.721401` on `dotnet/diagnostics:release/stable`, commit `3f576a7`). This is the latest stable release that contains both upstream fixes for the no-tty `SetCursorPosition(0, -1)` crash that occurs when `collect-linux` is spawned with redirected stdout: `dotnet/diagnostics#5771` ("Fix no-tty crash with console capability aware ProgressWriter") and `dotnet/diagnostics#5745` ("Handle redirected input/output in collect-linux"). 10.x is not on nuget.org; the agent writes a hermetic `NuGet.config` next to the tool path and resolves from the dnceng public `dotnet-tools` feed via `--configfile`.

## Stacking

PR A (dotnet/crank#886) has **merged** into `dotnet/crank:main` as squash commit `d392425`. This branch has been rebased onto current `main` and now contains only the collect-linux commits.

### DIFF
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -150,7 +150,6 @@ namespace Microsoft.Crank.Agent
         private static string _perfviewPath;
         private static string _ultraPath;
         private static string _dotnetTracePath;
-        private static readonly SemaphoreSlim _dotnetTraceInstallLock = new(1, 1);
         private static string _dotnetInstallPath;
         private static string _localUrl;
 
@@ -6488,97 +6487,90 @@ namespace Microsoft.Crank.Agent
         // ---- dotnet-trace CLI integration (collect / collect-linux modes) ----
 
         // Lazy installs the pinned dotnet-trace CLI under {_rootTempDir}/dotnet-trace-tools.
-        // Idempotent and safe to call concurrently across jobs (guarded by a semaphore).
+        // Idempotent: the `File.Exists` short-circuit makes repeat calls cheap, matching
+        // the lock-free pattern used by the SDK install path (EnsureDotnetInstallExistsAsync).
         // Post-install runs `dotnet-trace --version` and throws if the reported version
         // does not contain the pinned constant.
         private static async Task EnsureDotnetTraceCliInstalledAsync()
         {
-            await _dotnetTraceInstallLock.WaitAsync();
-            try
+            if (!string.IsNullOrEmpty(_dotnetTracePath) && File.Exists(_dotnetTracePath))
             {
-                if (!string.IsNullOrEmpty(_dotnetTracePath) && File.Exists(_dotnetTracePath))
-                {
-                    return;
-                }
+                return;
+            }
 
-                var toolPath = Path.Combine(_rootTempDir ?? Path.GetTempPath(), "dotnet-trace-tools");
-                Directory.CreateDirectory(toolPath);
+            var toolPath = Path.Combine(_rootTempDir ?? Path.GetTempPath(), "dotnet-trace-tools");
+            Directory.CreateDirectory(toolPath);
 
-                var exeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet-trace.exe" : "dotnet-trace";
-                var binPath = Path.Combine(toolPath, exeName);
+            var exeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet-trace.exe" : "dotnet-trace";
+            var binPath = Path.Combine(toolPath, exeName);
 
-                if (!File.Exists(binPath))
-                {
-                    var dotnetExe = GetDotNetExecutable(_dotnethome);
-                    Log.Info($"Installing dotnet-trace {DotnetTraceVersion} to {toolPath}");
+            if (!File.Exists(binPath))
+            {
+                var dotnetExe = GetDotNetExecutable(_dotnethome);
+                Log.Info($"Installing dotnet-trace {DotnetTraceVersion} to {toolPath}");
 
-                    // Write a hermetic NuGet.config alongside the tool so the
-                    // install resolves the dnceng dotnet-tools feed (which is
-                    // where 10.x previews live) regardless of any global
-                    // NuGet config the host might have. A bare `--add-source`
-                    // conflicts with package source mapping when the host has
-                    // it configured, so we pass a `--configfile` instead.
-                    var nugetConfigPath = Path.Combine(toolPath, "NuGet.config");
-                    File.WriteAllText(nugetConfigPath,
-                        "<?xml version=\"1.0\" encoding=\"utf-8\"?>" + Environment.NewLine +
-                        "<configuration>" + Environment.NewLine +
-                        "  <packageSources>" + Environment.NewLine +
-                        "    <clear />" + Environment.NewLine +
-                        "    <add key=\"dotnet-tools\" value=\"" + DotnetToolsFeedUrl + "\" />" + Environment.NewLine +
-                        "  </packageSources>" + Environment.NewLine +
-                        "</configuration>" + Environment.NewLine);
+                // Write a hermetic NuGet.config alongside the tool so the
+                // install resolves the dnceng dotnet-tools feed (which is
+                // where 10.x previews live) regardless of any global
+                // NuGet config the host might have. A bare `--add-source`
+                // conflicts with package source mapping when the host has
+                // it configured, so we pass a `--configfile` instead.
+                var nugetConfigPath = Path.Combine(toolPath, "NuGet.config");
+                File.WriteAllText(nugetConfigPath,
+                    "<?xml version=\"1.0\" encoding=\"utf-8\"?>" + Environment.NewLine +
+                    "<configuration>" + Environment.NewLine +
+                    "  <packageSources>" + Environment.NewLine +
+                    "    <clear />" + Environment.NewLine +
+                    "    <add key=\"dotnet-tools\" value=\"" + DotnetToolsFeedUrl + "\" />" + Environment.NewLine +
+                    "  </packageSources>" + Environment.NewLine +
+                    "</configuration>" + Environment.NewLine);
 
-                    var installArgs = $"tool install dotnet-trace --version {DotnetTraceVersion} --tool-path \"{toolPath}\" --configfile \"{nugetConfigPath}\"";
-                    var result = await ProcessUtil.RunAsync(
-                        dotnetExe,
-                        installArgs,
-                        log: true,
-                        throwOnError: false,
-                        captureOutput: true,
-                        captureError: true);
-
-                    if (result.ExitCode != 0 || !File.Exists(binPath))
-                    {
-                        throw new InvalidOperationException(
-                            $"`dotnet tool install dotnet-trace --version {DotnetTraceVersion}` failed (exit={result.ExitCode}). stderr: {result.StandardError?.Trim()}");
-                    }
-                }
-
-                var versionResult = await ProcessUtil.RunAsync(
-                    binPath, "--version",
-                    log: false,
+                var installArgs = $"tool install dotnet-trace --version {DotnetTraceVersion} --tool-path \"{toolPath}\" --configfile \"{nugetConfigPath}\"";
+                var result = await ProcessUtil.RunAsync(
+                    dotnetExe,
+                    installArgs,
+                    log: true,
                     throwOnError: false,
                     captureOutput: true,
                     captureError: true);
 
-                if (versionResult.ExitCode != 0)
+                if (result.ExitCode != 0 || !File.Exists(binPath))
                 {
                     throw new InvalidOperationException(
-                        $"`dotnet-trace --version` failed (exit={versionResult.ExitCode}). stderr: {versionResult.StandardError?.Trim()}");
+                        $"`dotnet tool install dotnet-trace --version {DotnetTraceVersion}` failed (exit={result.ExitCode}). stderr: {result.StandardError?.Trim()}");
                 }
-
-                var reported = (versionResult.StandardOutput ?? string.Empty).Trim();
-                Log.Info($"dotnet-trace installed: {reported}");
-                if (!reported.Contains(DotnetTraceVersion, StringComparison.Ordinal))
-                {
-                    // Hard fail: a pinned-version install with a mismatched runtime
-                    // version is a "we don't know what we have" situation. Usually
-                    // means the tool-path was pre-populated with a different
-                    // dotnet-trace by an earlier agent run on a different pin.
-                    // Reset the cached path so a manual cleanup of `toolPath` lets
-                    // the next attempt re-run the installer.
-                    _dotnetTracePath = null;
-                    throw new InvalidOperationException(
-                        $"dotnet-trace reported version '{reported}' does not contain the pinned '{DotnetTraceVersion}'. " +
-                        $"Delete '{toolPath}' and retry, or update DotnetTraceVersion to match.");
-                }
-
-                _dotnetTracePath = binPath;
             }
-            finally
+
+            var versionResult = await ProcessUtil.RunAsync(
+                binPath, "--version",
+                log: false,
+                throwOnError: false,
+                captureOutput: true,
+                captureError: true);
+
+            if (versionResult.ExitCode != 0)
             {
-                _dotnetTraceInstallLock.Release();
+                throw new InvalidOperationException(
+                    $"`dotnet-trace --version` failed (exit={versionResult.ExitCode}). stderr: {versionResult.StandardError?.Trim()}");
             }
+
+            var reported = (versionResult.StandardOutput ?? string.Empty).Trim();
+            Log.Info($"dotnet-trace installed: {reported}");
+            if (!reported.Contains(DotnetTraceVersion, StringComparison.Ordinal))
+            {
+                // Hard fail: a pinned-version install with a mismatched runtime
+                // version is a "we don't know what we have" situation. Usually
+                // means the tool-path was pre-populated with a different
+                // dotnet-trace by an earlier agent run on a different pin.
+                // Reset the cached path so a manual cleanup of `toolPath` lets
+                // the next attempt re-run the installer.
+                _dotnetTracePath = null;
+                throw new InvalidOperationException(
+                    $"dotnet-trace reported version '{reported}' does not contain the pinned '{DotnetTraceVersion}'. " +
+                    $"Delete '{toolPath}' and retry, or update DotnetTraceVersion to match.");
+            }
+
+            _dotnetTracePath = binPath;
         }
 
         // Static prereq check for `DotNetTraceCollectMode=collect-linux`.
--- a/src/Microsoft.Crank.Agent/Startup.cs
+++ b/src/Microsoft.Crank.Agent/Startup.cs
@@ -63,13 +63,31 @@ namespace Microsoft.Crank.Agent
 
         // dotnet-trace CLI version pinned for the `DotNetTraceCollectMode=collect`
         // and `DotNetTraceCollectMode=collect-linux` paths. Must:
-        //  - Exist on nuget.org (the agent's `dotnet tool install -g` reaches the
-        //    public feed, not crank's internal dnceng feeds).
-        //  - Ship a `linux-arm64` runtime (cobalt-hosted aarch64).
-        //  - Contain the `collect-linux` verb. (First added Oct 2025; this version
-        //    is the first stable nuget.org release that carries it.)
-        // When bumping, re-verify all three.
-        private const string DotnetTraceVersion = "9.0.661903";
+        //  - Exist on the dnceng public dotnet-tools feed (the agent passes
+        //    `--configfile` below; nuget.org tops out at 9.0.661903 today
+        //    and does NOT carry the 10.x line that ships `collect-linux`).
+        //  - Be a single-file managed tool packaged under `tools/net8.0/any/`
+        //    so it works on the cobalt-hosted aarch64 host (no native bits
+        //    needed; the verified package layout is RID-agnostic).
+        //  - Contain the `collect-linux` verb. (First added Oct 2025.)
+        //  - Contain dotnet/diagnostics#5771 ("Fix no-tty crash with console
+        //    capability aware ProgressWriter") AND #5745 ("Handle redirected
+        //    input/output in collect-linux"). Earlier builds crash with
+        //    `SetCursorPosition(0, -1)` when stdout is redirected (which is
+        //    always the case when crank spawns the tool).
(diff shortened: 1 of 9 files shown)


src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,533 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
             return failed ? -1 : 0;
         }
 
+        // ---- dotnet-trace CLI integration (collect / collect-linux modes) ----
+
+        // Lazy installs the pinned dotnet-trace CLI under {_rootTempDir}/dotnet-trace-tools.
+        // Idempotent and safe to call concurrently across jobs (guarded by a semaphore).
+        // Post-install runs `dotnet-trace --version` and warns if the reported version
+        // does not contain the pinned constant.
```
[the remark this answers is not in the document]
LoopedBard3: Good catch — fixed in 6a450ee. The comment now says `throws` (matching the `InvalidOperationException` thrown a few lines below).

test/Microsoft.Crank.UnitTests/DotNetTraceCollectLinuxPreflightTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            string raw, int major, int minor, int build)
+        {
+            var actual = Startup.ParseKernelVersion(raw);
+
+            Assert.NotNull(actual);
+            Assert.Equal(major, actual.Major);
+            Assert.Equal(minor, actual.Minor);
+            // System.Version.Build is -1 when only Major.Minor was supplied;
+            // ParseKernelVersion treats missing build as 0 for ordering.
+            Assert.True(actual.Build == build || (build == 0 && actual.Build <= 0));
```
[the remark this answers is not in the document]
LoopedBard3: Right — `ParseKernelVersion` always constructs `new Version(major, minor, patch)` with missing patch normalized to 0, so `Build` is never -1. Simplified the assertion to a straight `Assert.Equal(build, actual.Build)` and updated the comment accordingly in 6a450ee.

test/Microsoft.Crank.UnitTests/DotNetTraceCliNormalizationTests.cs
```diff
@@ -0,0 +1,298 @@
...
+            // Provider specs (containing ':') are not keyword expressions.
+            var input = "Microsoft-Windows-DotNETRuntime:0x4:5";
+            var actual = Startup.NormalizeClrEventExpression(input, out var notes);
+
+            Assert.Equal(input, actual);
+            Assert.Null(notes);
+        }
+
+        [Fact]
+        public void Normalize_LeavesUnknownKeywordsAlone()
```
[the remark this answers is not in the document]
LoopedBard3: Renamed to `Normalize_LeavesNonAliasedKeywordExpressionAlone` in 6a450ee, with a short comment clarifying that `gc+jit` is a valid CLR keyword expression that simply uses no aliases we rewrite.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -5538,7 +5568,55 @@ private static void StartDotNetTrace(Job job)
...
+                    break;
+
+                case "collect":
+                    dotnetTraceTask = CollectViaDotNetTraceCliAsync(
+                        job, dotnetTraceManualReset, "collect", new FileInfo(job.PerfViewTraceFile));
+                    break;
+
+                case "collect-linux":
+                {
+                    var (ok, err) = ValidateCollectLinuxPrerequisites();
```
DrewScoggins: Why are we using go style syntax for this function call? We should use more idiomatic C#

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -5538,7 +5568,55 @@ private static void StartDotNetTrace(Job job)
...
+                        dotnetTraceTask = Task.FromResult(-1);
+                        dotnetTraceManualReset.Set();
+                        break;
+                    }
+                    dotnetTraceTask = CollectViaDotNetTraceCliAsync(
+                        job, dotnetTraceManualReset, "collect-linux", new FileInfo(job.PerfViewTraceFile));
+                    break;
+                }
+
+                default:
```
DrewScoggins, line 5616: We should do this verification in the args parsing logic.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
 
+        // ---- dotnet-trace CLI integration (collect / collect-linux modes) ----
+
+        // Lazy installs the pinned dotnet-trace CLI under {_rootTempDir}/dotnet-trace-tools.
+        // Idempotent and safe to call concurrently across jobs (guarded by a semaphore).
+        // Post-install runs `dotnet-trace --version` and throws if the reported version
+        // does not contain the pinned constant.
+        private static async Task EnsureDotnetTraceCliInstalledAsync()
+        {
+            await _dotnetTraceInstallLock.WaitAsync();
```
DrewScoggins: Does this actually do anything? I don't think other copies of crank are going to be running, and if they were, I don't think they would share the same semaphore.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
+            try
+            {
+                if (!string.IsNullOrEmpty(_dotnetTracePath) && File.Exists(_dotnetTracePath))
```
DrewScoggins: Along the lines of the last verification comment, I think this should happen at the args parsing level. I am fairly certain we do that for other things.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                return (false, BuildCollectLinuxPrereqError(osDesc, euidText, kernelRaw));
+            }
+
+            uint euid;
+            try
+            {
+                euid = Mono.Unix.Native.Syscall.geteuid();
```
DrewScoggins, line 6595: Do we already have a dependency on this package? Seems like we might want to get rid of Mono where it makes sense.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
+                _dotnetTraceInstallLock.Release();
+            }
+        }
+
+        // Static prereq check for `DotNetTraceCollectMode=collect-linux`.
+        // Returns (true, null) on success, or (false, <verbatim error text>) on failure.
+        // The error text is contractual — see plan §"Container caveat for collect-linux
+        // prerequisites" for the verbatim form. **Callers must hard-fail the job on
+        // failure rather than silently substituting `collect` mode.**
+        internal static (bool ok, string error) ValidateCollectLinuxPrerequisites()
```
DrewScoggins, line 6581: Larger question. Why do we have this function? Won't the code is it possible for this function to fail validation, but we can call collect-linux and fail in some silent way?

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
+        // Returns null on unparseable input.
+        internal static Version ParseKernelVersion(string raw)
+        {
+            if (string.IsNullOrWhiteSpace(raw))
+            {
+                return null;
+            }
+
+            var trimmed = raw.Trim();
+            var stopIdx = trimmed.IndexOfAny(new[] { '-', '+', ' ', '~' });
```
DrewScoggins: Can we replace this with a regex? Also are there proven cases where we would need all of the other delimiters?

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Done in `ad0e2ea` -- pulled mode validation out of `StartDotNetTrace` (which only runs after the app launches) into a new `ValidateDotNetTraceOptions(job)` that runs in the `JobState.New` acceptance path, alongside our existing early checks (e.g. the IIS-on-Linux `NotSupported` gate). An invalid `DotNetTraceCollectMode` now fails the job before any restore/build/launch. This `default:` arm is kept as a defensive-only guard.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Removed in `8dc235a`. You're right -- it only serialized concurrent installs within a single agent process and couldn't coordinate across separate crank agent processes, which wouldn't share the `SemaphoreSlim` anyway. The SDK install path (`EnsureDotnetInstallExistsAsync`) uses no such lock, so this was inconsistent. Now relying on the `File.Exists` idempotency short-circuit, matching the established pattern.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Same fix as the mode-validation comment (`ad0e2ea`). The collect-linux host preflight (root, kernel >= 6.4) now runs at job acceptance via `ValidateDotNetTraceOptions` -> `ValidateCollectLinuxPrerequisites`, so an unsupported host fails immediately instead of after the app starts. `StartDotNetTrace`'s collect-linux arm no longer re-runs it.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: We already depend on `Mono.Unix` -- it's a `PackageReference` in `Microsoft.Crank.Agent.csproj` and is used today for `Syscall.kill(...SIGTERM/SIGINT)` in the process-stop paths (~lines 1810, 1824, 2101, 6923), so `geteuid()` here adds no new dependency. Agree that dropping Mono where it makes sense is worthwhile, but that's a repo-wide change to all the signal-sending code and feels out of scope for this PR -- happy to file a separate issue for it.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: The function fails fast with a clear, actionable message: without it, collect-linux on a non-root or old-kernel host fails deep inside the spawned `dotnet-trace` process with a raw error. It isn't silent today either -- if collect-linux fails at runtime, `CollectViaDotNetTraceCliAsync` captures the non-zero exit and sets `job.Error` -- but the preflight turns `benchmark ran, trace empty` into `job rejected up front: requires root + kernel >= 6.4, detected X`. With `ad0e2ea` it now runs at job acceptance, before the app launches.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -6366,6 +6444,532 @@ private static async Task<int> Collect(ManualResetEvent shouldExit, int processI
...
+                await Task.Delay(200);
+            }
+
+            // Benchmark signaled stop. Send SIGINT for a graceful flush+rundown.
+            Log.Info("Stopping dotnet-trace CLI (sending SIGINT)");
+            try
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    Mono.Unix.Native.Syscall.kill(proc.Id, Mono.Unix.Native.Signum.SIGINT);
```
DrewScoggins, line 6955: Same syscall question as previous comment.

LoopedBard3: Same as other syscall. We can take a look at removing this dependency, but I would do this in another PR.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: We can take a look at removing this dependency, but I would do this in another PR.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Done in `3d7cb21`. Replaced the `IndexOfAny` delimiter list with an anchored regex `^(\d+)\.(\d+)(?:\.(\d+))?` that captures the leading `major.minor[.patch]` and naturally stops at whatever suffix follows -- so we no longer enumerate delimiters at all. On the second question: in practice `uname -r` / `/proc/sys/kernel/osrelease` only ever uses `-` (Ubuntu `6.8.0-1018-azure`, `6.4.0-rc1`) and occasionally `+` (locally-built kernels). The `\u0020` (space) and `~` entries were defensive, not from proven cases. The regex makes that moot since it ignores everything after the numeric version regardless of delimiter -- I added test cases for `~` and space suffixes to lock that in.

src/Microsoft.Crank.Agent/Startup.cs
DrewScoggins: I suppose I am not fully convinced that this level of defensiveness is warranted for all of the extra complexity of the code.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -5591,31 +5605,57 @@ private static void StartDotNetTrace(Job job)
...
+        // Validates dotnet-trace options that can be checked at job-acceptance
+        // time -- before any asset restore, clone/build, or app launch -- so a
+        // misconfigured job fails fast instead of after a full startup. Returns
+        // (true, null) when there is nothing to collect or the options are valid,
+        // or (false, <error text>) when the job should be failed. For
+        // collect-linux this also runs the host prerequisite check (Linux, root,
+        // kernel >= 6.4); those are properties of the agent host, so checking
+        // them here is strictly earlier than -- and equivalent to -- the old
+        // check inside StartDotNetTrace.
+        internal static (bool ok, string error) ValidateDotNetTraceOptions(Job job)
```
DrewScoggins, line 5637: We still have the Go-ism here. Does this exist for a particular reason?

DrewScoggins (approved):
Still not convinced on the necessity of the euid validation method, and still think we should get rid of the Go-style tuple return, but those aren't pr stoppers.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: I am open to other levels of defensiveness, but I know with the perfcollect flow, it would only show as 404 when trying to transfer the trace back on the controller side. The agent logs did show that the perfcollect tool failed to install due to LTTng install issues, but that wasn't known on the controller and still resulted in the time being take for the test to run before we got to the error. I am not 100% sure this will hit with the dotnet-trace flow, so I can test that and go from there.

src/Microsoft.Crank.Agent/Startup.cs
LoopedBard3: Missed leaving this comment on the Go syntax. Basically this syntax is already being used a couple of times throughout the file, such as with DockerBuildAndRun, GetCurrentVersions, and ParseLatestVersionFile. I don't have a strong preference so we can discuss more and decide whether to follow up.

PR #896, opened 2026-09-21
Update dependencies, migrate to .NET 10, and enable Dependabot

sebastienros:
## Summary

- Update centrally managed dependencies to the latest stable versions available on the existing dnceng feeds. Keep Jint at 4.16.2 and MessagePack at 3.1.8 because their newer patches are not mirrored yet; do not add nuget.org.
- Migrate the .NET 8 projects and associated samples to .NET 10, pin SDK 10.0.401, and update documentation and container images. Preserve the EventSources library .NET Standard target. Move the default agent container to Ubuntu 24.04 and use matching package and debug-symbol repositories.
- Migrate the bots to stable System.CommandLine APIs, including explicit option binding, defaults, URI parsing, and argument splitting.
- Adapt JSON-schema validation to JsonElement and isolated registries, preserve repeated configuration loading, and report detailed validation errors.
- Update CsWin32 safe-handle calls, managed identity construction, and HTTP client PFX loading. Add Microsoft.Data.SqlClient.Extensions.Azure to preserve Entra authentication after the SqlClient upgrade.
- Replace obsolete agent WebHost and Windows service APIs with generic hosting. Remove redundant framework package references and migrate model certificate loading to X509CertificateLoader, with PFX/private-key, DER, PEM, and incorrect-password coverage.
- Configure public Windows/Linux and official Windows jobs with UseDotNet@2, selecting the SDK from global.json and installing into the Arcade .dotnet directory. This exports DOTNET_ROOT so native test apphosts find .NET 10 rather than the older machine-wide runtime.
- Add weekly Dependabot updates using the existing feeds, group minor/patch updates, and exclude shared Arcade infrastructure.

## Validation

Using .NET SDK 10.0.401 on macOS:

- Reproduced all seven Windows CI errors using the actual Arcade entry point and its warnings-as-errors policy. Earlier direct builds did not enforce that policy.
- After the fixes, `bash eng/common/build.sh --restore --build --pack --ci --configuration Release` succeeded with zero warnings and zero errors.
- Ran the full Arcade test target: 152 unit tests and 2 template tests passed. Integration results were 45 passed, 1 failed, and 2 expected platform skips. The sole failure was a native Bombardier process exiting with code 137 during MultiClients; that test passed on an isolated rerun. All 46 runnable integration cases therefore passed, but not in a single clean full-suite run.
- Solution and hello/post sample builds, controller configuration generation, and agent tool packaging succeeded.
- Dependabot configuration passed its published JSON schema; all 44 central package entries were checked against the existing public feed.
- Investigated the subsequent Windows integration failure: TestPrecommandWithVariables launched hello.exe against C:\Program Files\dotnet, which had no .NET 10 runtime. After configuring the pipeline runtime environment, both precommand integration tests passed locally with DOTNET_ROOT set to the checkout SDK directory.
- Documented the portable CI command in docs/development.md, including avoiding machine-wide cleanup on shared development machines.

Docker images were not built because the Docker daemon was unavailable. Windows service execution, real Azure Relay connectivity, and Linux-only tracing/dump behavior were not exercised on macOS. Fresh remote CI results are pending after the fixes.


src/Microsoft.Crank.RegressionBot/Program.cs
```diff
@@ -60,54 +59,50 @@ static Program()
...
+                new Option<long>("--install-id") { Description = "The GitHub installation id." },
+                new Option<string>("--connectionstring")
+                {
+                    Description = "The database connection string, or environment variable name containing it. (Secured)",
+                    Required = true
+                },
+                new Option<string[]>("--config")
+                {
+                    Description = "The path to a configuration file. (Can be repeated)",
+                    Required = true
```
LoopedBard3: Can we set `AllowMultipleArgumentsPerToken = true` here to preserve the previous parser behavior? `--config first.yml second.yml` was accepted before this upgrade, but the migrated command rejects `second.yml` as an unrecognized argument. Repeating `--config` still works, which is what the new tests cover.

I checked the public caller in [aspnet/Benchmarks/build/regressionbot.yml](https://github.com/aspnet/Benchmarks/blob/main/build/regressionbot.yml): it passes one configuration file per invocation, so that caller is not affected. This is a backward-compatibility concern rather than a confirmed break in that deployment. A test with two filenames after a single `--config` would cover it.

LoopedBard3 (changes requested):
Tried to run the docker build.sh on a machine and hit the following error:
```
 => [stage-1  2/14] RUN dotnet tool install -g dotnet-symbol                                                      17.7s
 => [build-env 3/3] RUN dotnet publish -c Release src/Microsoft.Crank.Agent --output /app --framework net10.0     27.0s
 => [stage-1  3/14] RUN apt-get update     && apt-get install -y --no-install-recommends         git         pro  58.7s
 => [stage-1  4/14] RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc &&     echo 2fa9c05d591a1582a9  5.7s
 => [stage-1  5/14] RUN apt-get update     && apt-get install -y --no-install-recommends         g++ make binuti  24.1s
 => [stage-1  6/14] RUN apt-get update     && apt-get install -y --no-install-recommends ubuntu-dbgsym-keyring     4.9s
 => ERROR [stage-1  7/14] RUN apt-get update     && apt-get install -y --no-install-recommends         openssl li  2.8s
------
 > [stage-1  7/14] RUN apt-get update     && apt-get install -y --no-install-recommends         openssl libssl-dev openssl-dbgsym libssl3t64-dbgsym     && openssl version:
0.383 Hit:1 https://packages.microsoft.com/ubuntu/24.04/prod noble InRelease
0.631 Hit:2 http://archive.ubuntu.com/ubuntu noble InRelease
0.692 Hit:3 http://archive.ubuntu.com/ubuntu noble-updates InRelease
0.763 Hit:4 http://archive.ubuntu.com/ubuntu noble-backports InRelease
0.945 Hit:5 http://security.ubuntu.com/ubuntu noble-security InRelease
1.207 Ign:6 https://ddebs.ubuntu.com noble InRelease
1.343 Ign:7 https://ddebs.ubuntu.com noble-updates InRelease
1.481 Err:8 https://ddebs.ubuntu.com noble Release
1.481   Redirection from https to 'http://ddebs.ubuntu.com/dists/noble/Release' is forbidden [IP: 185.125.190.48 443]
1.617 Err:9 https://ddebs.ubuntu.com noble-updates Release
1.617   Redirection from https to 'http://ddebs.ubuntu.com/dists/noble-updates/Release' is forbidden [IP: 185.125.190.48 443]
1.635 Reading package lists...
2.760 E: The repository 'https://ddebs.ubuntu.com noble Release' does not have a Release file.
2.760 E: The repository 'https://ddebs.ubuntu.com noble-updates Release' does not have a Release file.
------
Dockerfile:70
--------------------
  69 |
  70 | >>> RUN apt-get update \
  71 | >>>     && apt-get install -y --no-install-recommends \
  72 | >>>         openssl libssl-dev openssl-dbgsym libssl3t64-dbgsym \
  73 | >>>     && openssl version
  74 |
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c apt-get update     && apt-get install -y --no-install-recommends         openssl libssl-dev openssl-dbgsym libssl3t64-dbgsym     && openssl version" did not complete successfully: exit code: 100
```


From CoPilot:
```
The Docker build fails because the `https://ddebs.ubuntu.com` repositories redirect to HTTP, which APT rejects as a protocol downgrade. This causes `apt-get update` to exit with code 100 before installing the OpenSSL debug-symbol packages. The recommended fix is to change both the `noble` and `noble-updates` repository URLs to `http://ddebs.ubuntu.com`, while retaining the existing `signed-by=/usr/share/keyrings/ubuntu-dbgsym-keyring.gpg` configuration so APT continues verifying repository signatures.
```
Going to leave a comment specifically with its recommended fix.

docker/agent/Dockerfile
```diff
@@ -61,12 +61,14 @@ RUN apt-get update \
...
 
 ENV DEBIAN_FRONTEND=noninteractive
-# Add the Debian sid repository
-RUN echo 'deb http://deb.debian.org/debian sid main' >> /etc/apt/sources.list \
-    && echo 'deb http://deb.debian.org/debian-debug sid-debug main' >> /etc/apt/sources.list
+# Install matching OpenSSL debug symbols from the Ubuntu repositories.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends ubuntu-dbgsym-keyring \
+    && echo 'deb [signed-by=/usr/share/keyrings/ubuntu-dbgsym-keyring.gpg] https://ddebs.ubuntu.com noble main restricted universe multiverse' > /etc/apt/sources.list.d/ddebs.list \
+    && echo 'deb [signed-by=/usr/share/keyrings/ubuntu-dbgsym-keyring.gpg] https://ddebs.ubuntu.com noble-updates main restricted universe multiverse' >> /etc/apt/sources.list.d/ddebs.list
```
LoopedBard3, line 68: The HTTPS endpoint redirects to HTTP, which APT rejects, causing `apt-get update` to fail. Use HTTP directly for both suites while retaining the existing repository-signature verification.

```suggestion
    && echo 'deb [signed-by=/usr/share/keyrings/ubuntu-dbgsym-keyring.gpg] http://ddebs.ubuntu.com noble main restricted universe multiverse' > /etc/apt/sources.list.d/ddebs.list \
    && echo 'deb [signed-by=/usr/share/keyrings/ubuntu-dbgsym-keyring.gpg] http://ddebs.ubuntu.com noble-updates main restricted universe multiverse' >> /etc/apt/sources.list.d/ddebs.list
```

LoopedBard3: This is one option suggested by CoPilot locally.

src/Microsoft.Crank.Agent/Startup.cs
```diff
@@ -532,10 +529,25 @@ private static async Task<int> Run(string url, string hostname, string dockerHos
...
+            var builder = new HostBuilder()
+                .ConfigureWebHost(webHostBuilder => ConfigureWebHost(webHostBuilder, url));
+
+            if (_runAsService.HasValue() && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                builder.UseWindowsService();
+            }
 
-            var host = builder.Build();
+            using var host = builder.Build();
```
LoopedBard3, line 545: A local agent review found that when using --relay, this will hit an issue due to "disposing the host calls `CompositeServer.Dispose()` after the DI provider has been marked disposed. Its `_servers` collection is a deferred query that calls `GetService(...)`, so the `foreach` throws `ObjectDisposedException` while enumerating it". It recommends updating the CompositeServer.Dispose() a no-op (in src/Microsoft.Crank.Agent/CompositeRelayServer.cs) and adding a regression test that resolves the combined relay/HTTP server and disposes the host, something like:
<details>
<summary>Proposed Regression Test</summary>

```csharp
using System.Linq;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Hosting.Server;
using Microsoft.Crank.Agent;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Xunit;

namespace Microsoft.Crank.UnitTests;

public class CompositeServerTests
{
    [Fact]
    public void CombinedRelayAndHttpHostDisposesWithoutThrowing()
    {
        using var host = new HostBuilder()
            .ConfigureWebHost(web =>
            {
                web.UseKestrel();
                web.UseAzureRelay(options => options.UrlPrefixes.Add(
                    "Endpoint=sb://unused.invalid/;EntityPath=test;" +
                    "SharedAccessKeyName=test;" +
                    "SharedAccessKey=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="));
                web.Configure(_ => { });
                web.ConfigureServices(services =>
                {
                    var descriptors = services.Where(d =>
                        d.Lifetime == ServiceLifetime.Singleton &&
                        typeof(IServer).IsAssignableFrom(d.ServiceType))
                        .ToArray();

                    foreach (var descriptor in descriptors)
                    {
                        services.Remove(descriptor);
                        services.AddSingleton(descriptor.ImplementationType);
                    }

                    services.AddSingleton<IServer>(provider =>
                        new CompositeServer(descriptors.Select(d =>
                            (IServer)provider.GetRequiredService(d.ImplementationType))));
                });
            })
            .Build();

        Assert.IsType<CompositeServer>(host.Services.GetRequiredService<IServer>());
    }
}
``` 

</details>

sebastienros: Confirmed and fixed in d9e66da. The new Kestrel + Azure Relay host regression test reproduced `ObjectDisposedException` from `CompositeServer.Dispose()` resolving services after the provider was disposed. `Dispose()` is now a no-op because DI owns the child servers. Added a second test proving the composite does not dispose its children and the provider disposes each exactly once. Both tests failed before the fix and pass afterward on .NET 10 with warnings treated as errors. This applies to the combined `--relay` / `--relay-enable-http` path.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

docker/agent/Dockerfile
sebastienros: Confirmed and fixed in d9e66da. Both `https://ddebs.ubuntu.com/dists/noble/InRelease` and the `noble-updates` URL return HTTP 301 redirects to HTTP; the direct HTTP URLs return 200. Switched both suites to HTTP, retaining `ubuntu-dbgsym-keyring` and the explicit `signed-by` checks. Full Docker build validation wasn't available because the local Docker daemon is not running.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

LoopedBard3:
The image now builds, but Relay startup fails with `MissingMethodException` for `HybridConnectionListener(Uri, TokenProvider)` when using `--relay --relay-enable-http`.

The existing `Microsoft.Azure.Relay.AspNetCore` 1.3.15173 binary calls that two-argument constructor. I inspected the published NuGet assemblies: `Microsoft.Azure.Relay` **3.0.1** contains it, while **3.1.1** replaces it with `(Uri, TokenProvider, bool = true)`. The optional argument preserves source compatibility when recompiling, not compatibility with the already-built adapter. The upstream change is https://github.com/Azure/azure-relay-dotnet/pull/181.

I reproduced the failure with the current agent, then rebuilt with **only `Microsoft.Azure.Relay` reverted to 3.0.1**. The missing-method exception disappeared and startup reached the network connection attempt (which intentionally failed against a dummy `.invalid` endpoint). This was not an authenticated live-Relay test.

**Could we revert just the `Microsoft.Azure.Relay` entry in `Directory.Packages.props` from 3.1.1 to 3.0.1 for now, leaving the other updates intact?** Rebuild the agent/image rather than replacing the DLL in an existing output. We can track restoration of the constructor overload upstream and migration off the archived ASP.NET adapter separately.

Directory.Packages.props
```diff
@@ -4,51 +4,52 @@
...
-    <PackageVersion Include="Microsoft.Azure.Relay" Version="3.0.1" />
+    <PackageVersion Include="Azure.Identity" Version="1.21.0" />
+    <PackageVersion Include="Azure.Messaging.ServiceBus" Version="7.20.2" />
+    <PackageVersion Include="Fluid.Core" Version="2.40.0" />
+    <PackageVersion Include="Jint" Version="4.16.2" />
+    <PackageVersion Include="JsonSchema.Net" Version="9.4.0" />
+    <PackageVersion Include="McMaster.Extensions.CommandLineUtils" Version="5.1.0" />
+    <PackageVersion Include="MessagePack" Version="3.1.8" />
+    <PackageVersion Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="10.0.12" />
+    <PackageVersion Include="Microsoft.Azure.Relay" Version="3.1.1" />
```
LoopedBard3: The image now builds, but Relay startup fails with `MissingMethodException` for `HybridConnectionListener(Uri, TokenProvider)` when using `--relay --relay-enable-http`.

The existing `Microsoft.Azure.Relay.AspNetCore` 1.3.15173 binary calls that two-argument constructor. I inspected the published NuGet assemblies: `Microsoft.Azure.Relay` **3.0.1** contains it, while **3.1.1** replaces it with `(Uri, TokenProvider, bool = true)`. The optional argument preserves source compatibility when recompiling, not compatibility with the already-built adapter. The upstream change is https://github.com/Azure/azure-relay-dotnet/pull/181.

I reproduced the failure with the current agent, then rebuilt with **only `Microsoft.Azure.Relay` reverted to 3.0.1**. The missing-method exception disappeared and startup reached the network connection attempt (which intentionally failed against a dummy `.invalid` endpoint). This was not an authenticated live-Relay test.

**Could we revert just the `Microsoft.Azure.Relay` entry in `Directory.Packages.props` from 3.1.1 to 3.0.1 for now, leaving the other updates intact?** Rebuild the agent/image rather than replacing the DLL in an existing output. We can track restoration of the constructor overload upstream and migration off the archived ASP.NET adapter separately.
```suggestion
    <PackageVersion Include="Microsoft.Azure.Relay" Version="3.0.1" />
```

sebastienros: Confirmed; this is a binary-compatibility regression, not a false positive. I inspected both published Relay API surfaces and reproduced the exact `MissingMethodException` in the current agent with `--relay --relay-enable-http`.

Fixed in af36774 by reverting only `Microsoft.Azure.Relay` to 3.0.1; all other dependency versions remain unchanged. Added a comment documenting the constraint and a regression test requiring the exact `(Uri, TokenProvider)` constructor. That test fails with 3.1.1 and passes with 3.0.1; all 11 targeted dependency/composite-server tests pass with warnings treated as errors.

Restored, rebuilt, and published the agent to a fresh output directory rather than replacing a DLL. Its dependency manifest resolves Relay 3.0.1, and startup now reaches the expected DNS/WebSocket failure against `unused.invalid`, without `MissingMethodException`. This was not an authenticated live-Relay test. I could not rebuild the Docker image because the local Docker daemon is not running.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

LoopedBard3 (approved):
This looks good to me. Verified that the dockerfile builds and runs (with /jobs api responding as expected with no jobs) so I think we should merge this.

